PR #10, opened 2020-12-02
BL-583 Add DistanceFromScreen to InputActions

MattGrayUL:
Adjusted VirtualScreen to use distance to the screen as opposed to a virtual screen plane at 5cm from the screen. 

Added DistanceFromScreen to InputActions across both sides of the connection.

Fixed a bug where the WSReceiverQueue would output default inputActions if none were in the queue.

### DIFF
--- a/ScreenControl/Assets/ScreenControl/Client/Scripts/CoreConnection/WebSocketReceiverQueue.cs
+++ b/ScreenControl/Assets/ScreenControl/Client/Scripts/CoreConnection/WebSocketReceiverQueue.cs
@@ -19,19 +19,25 @@ namespace Ultraleap.ScreenControl.Client
             ClientInputAction action;
             while (receiveQueue.Count > queueCullLimit)
             {
-                receiveQueue.TryDequeue(out action);
-
-                // Stop skipping through the queue this frame if we have just handled a 'key' input event
-                if (action.InputType != InputType.MOVE)
+                if (receiveQueue.TryPeek(out action))
                 {
-                    coreConnection.HandleInputAction(action);
-                    return;
+                    receiveQueue.TryDequeue(out action);
+
+                    // Stop skipping through the queue this frame if we have just handled a 'key' input event
+                    if (action.InputType != InputType.MOVE)
+                    {
+                        coreConnection.HandleInputAction(action);
+                        return;
+                    }
                 }
             }
 
-            // Parse newly received messages
-            receiveQueue.TryDequeue(out action);
-            coreConnection.HandleInputAction(action);
+            if (receiveQueue.TryPeek(out action))
+            {
+                // Parse newly received messages
+                receiveQueue.TryDequeue(out action);
+                coreConnection.HandleInputAction(action);
+            }
         }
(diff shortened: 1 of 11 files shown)


ScreenControl/Assets/ScreenControl/Client/Scripts/CoreConnection/WebSocketReceiverQueue.cs
```diff
@@ -19,19 +19,25 @@ void Update()
             ClientInputAction action;
             while (receiveQueue.Count > queueCullLimit)
             {
-                receiveQueue.TryDequeue(out action);
-
-                // Stop skipping through the queue this frame if we have just handled a 'key' input event
-                if (action.InputType != InputType.MOVE)
+                if (receiveQueue.TryPeek(out action))
```
NatGreenwood-UL, line 22: Do we need this peek? Isn't testing that there's something there covered by the content of the `while`?

ScreenControl/Assets/ScreenControl/Core/Scripts/Interactions/InteractionModules/GrabInteraction.cs
```diff
@@ -216,7 +216,7 @@ private void OnDrawGizmos()
                 Gizmos.color = Color.red;
                 Gizmos.DrawWireSphere(-GlobalSettings.virtualScreen.PhysicalScreenPlane.normal * GlobalSettings.virtualScreen.PhysicalScreenPlane.distance, 0.01f);
                 Gizmos.color = Color.green;
-                Gizmos.DrawWireSphere(planeHit_debug + (-GlobalSettings.virtualScreen.VirtualScreenPlane.normal * GlobalSettings.virtualScreen.VirtualScreenPlane.distance), 0.01f);
+                //Gizmos.DrawWireSphere(planeHit_debug + (-GlobalSettings.virtualScreen.VirtualScreenPlane.normal * GlobalSettings.virtualScreen.VirtualScreenPlane.distance), 0.01f);
```
NatGreenwood-UL, line 219: Why comment this out instead of delete it?

ScreenControl/Assets/ScreenControl/Client/Scripts/CoreConnection/WebSocketReceiverQueue.cs
MattGrayUL: The content of the while won't have run if we are at the bottom of the method and the condition of the while was never met

ScreenControl/Assets/ScreenControl/Core/Scripts/Interactions/InteractionModules/GrabInteraction.cs
MattGrayUL: If we wanted to keep the gizmos in the examples in future, we may want a reference for the gizmo that I have broken by deleting 'GlobalSettings.virtualScreen.VirtualScreenPlane'

NatGreenwood-UL (approved):

PR #24, opened 2021-02-12
BL-580: Update API Version Handshake for web compatibility

NatGreenwood-UL:


ScreenControl_Unity/Assets/ScreenControl/Client/Scripts/Connection/ScreenControlServiceTypes.cs
```diff
@@ -16,7 +16,9 @@ internal enum ActionCode
         CONFIGURATION_STATE,
         CONFIGURATION_RESPONSE,
         SET_CONFIGURATION_STATE,
-        REQUEST_CONFIGURATION_STATE
+        REQUEST_CONFIGURATION_STATE,
+        VERSION_HANDSHAKE,
+        VERSION_HANDSHAKE_RESPONSE
```
MattGrayUL, line 21: No docs to explain these

ScreenControl_Unity/Assets/ScreenControl/Client/Scripts/Connection/ServiceConnection.cs
```diff
@@ -19,6 +19,11 @@ public class ServiceConnection
         // A reference to the websocket we are connected to.
         WebSocket webSocket;
 
+        // Variable: handshakeCompleted
+        // Used internally in this class to know if the Version compatibility handshake with
+        // the server has successfully completed.
```
MattGrayUL, line 24: Do we need docs on private variables?

ScreenControl_Unity/Assets/ScreenControl/Client/Scripts/Connection/ServiceConnection.cs
```diff
@@ -27,18 +32,49 @@ public class ServiceConnection
         // use and redirects incoming messages to <OnMessage>.
```
MattGrayUL: Can we explain the handshake in the docs here?

ScreenControl_Unity/Assets/ScreenControl/Service/ScreenControlWsBehaviour.cs
```diff
@@ -39,50 +62,36 @@ public void SendConfigurationResponse(ConfigResponse _response)
...
+            }
+
+            else if (clientVersionParsed.Minor < _coreVersion.Minor)
+            {
+                return Compatibility.CLIENT_OUTDATED;
+            }
+            else if (clientVersionParsed.Minor > _coreVersion.Minor)
             {
                 return Compatibility.SERVICE_OUTDATED;
             }
```
MattGrayUL: Can these ifs be 'OR'-ed so we don't have duplicated returns?

ScreenControl_Unity/Assets/ScreenControl/Service/ScreenControlWsBehaviour.cs
```diff
@@ -124,5 +142,76 @@ protected override void OnMessage(MessageEventArgs _message)
...
+                response.message = "Request Rejected: Requests cannot be processed until handshaking is complete.";
+                Debug.LogError("Request Rejected: Requests cannot be processed until handshaking is complete.");
+                SendHandshakeResponse(response);
+                return;
+            }
+
+            if (!contentObj.ContainsKey(VersionInfo.API_HEADER_NAME))
+            {
+                // Send back immediate error
+                // "Request made before Handshake completed"
```
MattGrayUL: Comment is wrong

ScreenControl_Unity/Assets/ScreenControl/Service/WebSocketClientConnection.cs
```diff
@@ -53,16 +53,16 @@ private void InitialiseServer()
         {
             websocketInitalised = false;
 
+            receiverQueue = GetComponent<WebSocketReceiver>();
+            receiverQueue.SetWSClientConnection(Instance);
```
MattGrayUL: This is not required if we are a singleton.

ScreenControl_Unity/Assets/ScreenControl/Service/WebSocketClientConnection.cs
```diff
@@ -26,25 +27,24 @@ private void Awake()
...
+            InteractionManager.HandleInputAction -= Instance.SendInputActionToWebsocket;
         }
 
         private void SetupConnection(ScreenControlWsBehaviour behaviour)
         {
             if (behaviour != null)
             {
                 socketBehaviour = behaviour;
-                socketBehaviour.clientConnection = this;
+                socketBehaviour.clientConnection = Instance;
```
MattGrayUL: This is not required if we are a singleton?

ScreenControl_Web/src/Connection/ServiceConnection.ts
```diff
@@ -22,16 +24,42 @@ export class ServiceConnection {
     // A reference to the websocket we are connected to.
     webSocket: WebSocket;
 
+    // Variable: handshakeCompleted
+    // Used internally in this class to know if the Version compatibility handshake with the
+    // server has successfully completed.
```
MattGrayUL: Do we need to document private variables?

ScreenControl_Web/src/Connection/ServiceConnection.ts
```diff
@@ -22,16 +24,42 @@ export class ServiceConnection {
...
+    private handshakeCompleted: boolean;
+
     // Group: Functions
 
     // Function: constructor
     // The constructor for <ServiceConnection> that can be given a different IP Address and Port
     // to connect to on construction. This constructor also sets up the redirects of incoming
     // messages to <OnMessage>.
```
MattGrayUL: These docs should include handshake description

ScreenControl_Web/src/Connection/ServiceConnection.ts
```diff
@@ -42,14 +70,31 @@ export class ServiceConnection {
         }
     }
 
+    // Function: ConnectionResultCallback
+    // Passed into <SendMessage> as part of connecting to ScreenControl Service, handles the
+    // result of the Version Checking handshake.
+    private ConnectionResultCallback(response: WebSocketResponse): void {
+        console.log("Got a response from connection");
```
MattGrayUL: Do we need this log if both other cases also log the result?

MattGrayUL (changes requested):

ScreenControl_Unity/Assets/ScreenControl/Service/ScreenControlWsBehaviour.cs
NatGreenwood-UL: I moved from the original state to the new state because it wasn't prioritising the major versions correctly, so no, I think we need to do these individually

The issue came with comparing something like Service "0.9.1" and Client "1.0.1", because it first checks if the client is older _in either way_, it was returning an **older** client for the above because of the minor versions: 0<9.

Ideally we'd use the built-in comparisons in the Version type but because we don't care about the Hotfix version, and there's no way to ask those compares to ignore the Hotfix version, it would _always_ tell us they were incompatible

ScreenControl_Unity/Assets/ScreenControl/Service/WebSocketClientConnection.cs
NatGreenwood-UL: We actually aren't using that reference once it's passed any more so I'm just gonna strip it

MattGrayUL (approved):

PR #25, opened 2021-02-16
Feature/bl-578_web_client_cursor

MattGrayUL:


ScreenControl_Web/examples/TestExample.html
```diff
@@ -13,28 +13,28 @@
...
+    var newCursor2 = document.createElement('img');
+    newCursor2.src = "../images/Ring.png";
+    newCursor2.style.position = "absolute";
+    newCursor2.alt = "na";
+    newCursor2.width = 50;
+    newCursor2.height = 50;
+    newCursor2.style.zIndex = "1000";
+    document.body.appendChild(newCursor2);
+
+    var newCursor = document.createElement('img');
```
NatGreenwood-UL: Might be nice to make the second image embedded in the raw HTML and simply find it in code, to show off how users might do the same

ScreenControl_Web/src/Cursors/DotCursor.ts
```diff
@@ -0,0 +1,180 @@
...
+} from '../ScreenControlTypes';
+import { TouchlessCursor } from './TouchlessCursor';
+import { MapRangeToRange } from '../Utilities';
+
+ //Class: DotCursor
+ //This is an example Touchless Cursor which positions a dot on the screen at the hand location,
+ //and reacts to the current ProgressToClick of the action (what determines this depends on the
+ //currently active interaction).
+export class DotCursor extends TouchlessCursor{
+    readonly animationUpdateDuration: number = (1 / 30) * 1000;
```
NatGreenwood-UL, line 15: Might be worth chucking a comment in for this even if it's not docs-style just for us / externals if they decide they want to change the update rate

ScreenControl_Web/src/Cursors/DotCursor.ts
```diff
@@ -0,0 +1,180 @@
...
+                }
+                break;
+
+            case InputType.CANCEL:
+                break;
+        }
+    }
+
+    // Function: ShrinkCursor
+    // Shrinks the cursor to half of its original size over time set via the <constructor>.
```
NatGreenwood-UL: This is a little hard to parse, could we make it "over _the_ time passed into the contructor" or something like that? Exactly what is set in the constructor is a little lost IMO

ScreenControl_Web/src/Cursors/DotCursor.ts
```diff
@@ -0,0 +1,180 @@
...
+        this.cursorStartSize = [_cursor.clientWidth.valueOf(), _cursor.clientHeight.valueOf()];
+
+        this.animationSpeed[0] = (this.cursorStartSize[0] / 2) / (_animationDuration * 30);
+        this.animationSpeed[1] = (this.cursorStartSize[1] / 2) / (_animationDuration * 30);
+    }
+
+    // Function: UpdateCursor
+    // Used to update the cursor when recieving a "MOVE" <ClientInputAction>. Updates the
+    // cursor's position, as well as the size of the ring based on the current ProgressToClick.
+    UpdateCursor(_inputAction: Array<number> | ClientInputAction): void{
```
NatGreenwood-UL: Seeing as this function _can't_ actually take just an array of numbers, is there any way we can make this just take a CIA? Feels ugly to be able to take the type but not handle it if provided.

I think I'd rather change the base class to take a CIA and just use the position of it 🤔

ScreenControl_Web/src/Cursors/TouchlessCursor.ts
```diff
@@ -0,0 +1,46 @@
...
+    // Group: Variables
+
+    // Variable: cursor
+    // The HTMLElement that represents this cursor
+    cursor: HTMLElement;
+
+    // Group: Functions
+
+    // Function: constructor
+    // Initialises the cursor at its default state.
```
NatGreenwood-UL: Is this true here?  We don't touch the cursor in this function

NatGreenwood-UL (changes requested):

NatGreenwood-UL (approved):

PR #27, opened 2021-02-18
Feature/bl 579 web client configuration management

MattGrayUL:


ScreenControl_Web/src/Configuration/ConfigurationTypes.ts
```diff
@@ -0,0 +1,83 @@
+// Class: InteractionConfig
+// This class is a container for all of the settings related to the interactions being processed
+// by the ScreenControl Service. The settings at the root of this object will affect all
+// sensations. There are also some settings specific to the Hover and Hold interaction which can
+//  be modified by changing the contained <HoverAndHoldInteractionSettings>.
```
NatGreenwood-UL: Doublespaced

ScreenControl_Web/src/Configuration/ConfigurationTypes.ts
```diff
@@ -0,0 +1,83 @@
...
+// not modify a member of this class, its value will not change when the instance is sent to
+// ScreenControl Service.
+export interface PhysicalConfig {
+    // Property: ScreenHeightM
+    // The height of the screen in meters. This is needed in order to determine the relationship
+    // between hand location in the real world and pixel locations on screen.
+    ScreenHeightM: number;
+
+    // Property: LeapPositionRelativeToScreenBottomM
+    // The position (measured in meters) in 3d space of the Leap Motion camera relative to the
```
NatGreenwood-UL, line 70: Do we need to say something about the handedness here? For those who aren't exactly clear on how our coord syatems relate

ScreenControl_Web/src/Connection/ScreenControlServiceTypes.ts
```diff
@@ -31,22 +31,25 @@ export enum Compatibility {
...
-// // Class: ConfigStateResponse
-// // This is the structure of data received when requesting the current state of the configuration files
-// // from the Service.
-// export class ConfigStateResponse
-// {
-//    requestID: string;
-//    interaction: InteractionConfig;
-//    physical: PhysicalConfig;
+// Class: ConfigState
+// This is the structure of data received when requesting the current state of the configuration files
```
NatGreenwood-UL: Is this the structure _recieved in response to the request_ or the structure _used to make_ the request? I think this comment was written for the former but this structure is now the latter?
It seems it may be both in which case we should maybe generalise this rather than talking about requests/responses and just say "this is the structure we use to represent the config state of the Service"?

NatGreenwood-UL (commented):
Couple of small things worth checking over, but changes not _needed_.

NatGreenwood-UL (approved):

PR #29, opened 2021-03-01
BL-672: Support multiple clients connected simultaneously

NatGreenwood-UL:
This also reintroduces the cursor in the Service, which will be disabled during auto-config to prevent confusion.

### DIFF
--- a/ScreenControl_Unity/Assets/ScreenControl/Service/Scenes/ScreenControlService.unity
+++ b/ScreenControl_Unity/Assets/ScreenControl/Service/Scenes/ScreenControlService.unity
@@ -191,7 +191,7 @@ PrefabInstance:
     - target: {fileID: 2213489624516412116, guid: 157025b17471e7c41a7d20e7f28273b5,
         type: 3}
       propertyPath: m_IsActive
-      value: 0
+      value: 1
       objectReference: {fileID: 0}
     - target: {fileID: 2383233282598257216, guid: 157025b17471e7c41a7d20e7f28273b5,
         type: 3}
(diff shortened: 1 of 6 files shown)


ScreenControl_Unity/Assets/ScreenControl/Service/Scenes/ScreenControlService.unity
```diff
@@ -191,7 +191,7 @@ PrefabInstance:
     - target: {fileID: 2213489624516412116, guid: 157025b17471e7c41a7d20e7f28273b5,
```
MattGrayUL, line 191: Can some of these scene changes be prefab changes instead?

MattGrayUL (changes requested):
Nice and tidy, just noticed some scene changes that could be prefab changes to avoid confusion in future

ScreenControl_Unity/Assets/ScreenControl/Service/Scenes/ScreenControlService.unity
NatGreenwood-UL: Ok so just looking over the changes:

* The Auto-config screen gets a reference to the cursor on the Client prefab. Because this is across prefabs, it has to be scene-based
* ScreenControl Service prefab gets a "disableOnFocusLost" script, which needs a reference to the Screen Control Client to disable/enable it on focus (because it can't go on the Client prefab because then it'd be disabled when focus is lost... and never come back.) I think I'd rather leave this off the prefab because it's not a general use-case and is specific to the Service.

MattGrayUL: Okay that sounds fine then 👍

MattGrayUL (approved):

PR #32, opened 2021-03-04
BL-674: CI signs Service exe

NatGreenwood-UL:


ScreenControl_Unity/Assets/ScreenControl/Service/Version.txt
```diff
@@ -0,0 +1,3 @@
+ScreenControl Service Version: {SOFTWARE_VERSION}
+ScreenControl API Version: {API_VERSION}
+Release Reference: {RELEASE_REF}
```
MattGrayUL, line 3: Is this actually being used? I cannot see a copy of this with the build artifact

.gitlab-ci.yml
```diff
@@ -34,15 +37,50 @@ SC_Client::Unity::build:
...
+    - cp -r ${env:SC_CFW_PATH}/build/dist ./Client_Release/
+  artifacts:
+    name: "Screen_Control_Client_for_Web_${env:SC_CFU_VERSION}_${env:CI_COMMIT_SHA}"
+    paths:
+      - ./Client_Release/*
+    expire_in: 2 weeks
+    when: always
+  tags:
+    - unity
+    - win10
```
MattGrayUL: This refers to the web client, id that correct?

NatGreenwood-UL: These tags indicate which build machines the job can work on, and the Unity runners already have Node installed on them so this works fine 👍

ScreenControl_Unity/Assets/ScreenControl/Service/Version.txt
NatGreenwood-UL: Line 73 of the CI is supposed to copy it into the package, so I'll try and fix this 🤔

.gitlab-ci.yml
MattGrayUL: Sorry I should have been more clear, this branch/work is related to the Service, why is there a large chunk of web-specific changes?

NatGreenwood-UL: Ohhhhhh. Because I extended the Web branch to do this and the web branch hasn't yet been merged 😅

ScreenControl_Unity/Assets/ScreenControl/Service/Version.txt
NatGreenwood-UL: Fixed this in the latest commit, tis in the package now

MattGrayUL (approved):

PR #33, opened 2021-03-04
Feature/bl 586 web quick start example

MattGrayUL:


NatGreenwood-UL (approved):
Looks good to me, am slightly encouraged that clicking & dragging the regions with a mouse doesn't work either so I wouldn't _necessarily_ expect them to work with the inputs either.

PR #39, opened 2021-03-23
[BL-698] Add fading to DotCursor upon Cancel events

MattGrayUL:


ScreenControl_Web/src/Cursors/DotCursor.ts
```diff
@@ -106,8 +109,13 @@ export class DotCursor extends TouchlessCursor{
                 break;
 
             case InputType.CANCEL:
+                this.HideCursor();
```
NatGreenwood-UL, line 115: Do we want to mention this in the docs for the function?

NatGreenwood-UL (approved):
One thing might want a comment but otherwise good to go!

PR #41, opened 2021-03-23
BL-699: Service UI no longer responds to SC Input when minimised / alt-tabbed

NatGreenwood-UL:


MattGrayUL (changes requested):
Was the fix to add a reference in the inspector? I can't see what would solve the issue.

This PR also undoes a series of Unity version changes on develop so the version-based changes should be removed

NatGreenwood-UL:
After the accidental develop issues this has become entirely broken. Needs reimplementation. Closing MR, will open new branch.

PR #46, opened 2021-03-29
Bl-706: SC Service builds include licenses dir

NatGreenwood-UL:


MattGrayUL (commented):
I think i may be best to provide the full text of all licenses, so the OFL file would need to be updated

MattGrayUL (approved):

PR #48, opened 2021-03-30
Feature/bl 657 service UI tray icon

MattGrayUL:


NatGreenwood-UL (commented):
Tray Icon looking good, not sure if you want to merge this before/after the fix for the changing settings in the UI making it to the Service. 👍

PR #50, opened 2021-03-30
Feature/bl 655 windows service config updates

MattGrayUL:


NatGreenwood-UL (approved):
Looks good, tested locally and the UI changes now make it to Service, which also made it to a Web Client 👌

PR #52, opened 2021-04-01
Feature/bl 704 split service and UI scenes

MattGrayUL:


ScreenControl_Unity/Assets/ScreenControl/Service/Scripts.meta
```diff
@@ -1,5 +1,5 @@
 fileFormatVersion: 2
```
NatGreenwood-UL, line 1: Examples/Scripts dir becoming Service/Scripts folder seems odd? Guess it doesn't matter too much because noone will be importing / upgrading Service since the last release

ScreenControl_Unity/Assets/ScreenControl/ServiceUI/Scripts/Configuration/ConfigUI/GlobalSettingsConfigUI.cs
```diff
@@ -195,7 +195,6 @@ protected override void ValidateValues()
             {
                 ConfigManager.InteractionConfig.InteractionType = ScreenControlTypes.InteractionType.HOVER;
             }
-            InteractionManager.Instance.SetActiveInteractions(ConfigManager.InteractionConfig.InteractionType);
```
NatGreenwood-UL, line 198: Why aren't we setting this any more? :S

NatGreenwood-UL (commented):
Can we remove the _ prefix from the `UI Assets` dir?

ScreenControl_Unity/Assets/ScreenControl/Service/Scripts.meta
MattGrayUL: This is git reading what I did incorrectly perhaps? I deleted the Examples folder first and then created a new folder for Service

ScreenControl_Unity/Assets/ScreenControl/ServiceUI/Scripts/Configuration/ConfigUI/GlobalSettingsConfigUI.cs
MattGrayUL: The UI application doesn't have an InteractionManager anymore so we don't need to set it, the configuration file is still set and updated (which triggers SetActiveInteractions to be called on the _actual_ service via file watching)

NatGreenwood-UL: Ahhhhhh I see, that's all good then 👍

ScreenControl_Unity/Assets/ScreenControl/Service/Scripts.meta
NatGreenwood-UL: Ah yeah that's all good, as long as the GUIDs are different (they are) I don't think it'll cause issues

NatGreenwood-UL (approved):

PR #53, opened 2021-04-06
Feature/bl 601 clients request config states

MattGrayUL:


ScreenControl_Unity/Assets/ScreenControl/Client/Scripts/Connection/ServiceConnection.cs
```diff
@@ -145,5 +145,39 @@ internal void SendMessage(string _message, string _requestID, Action<WebSocketRe
 
             webSocket.Send(_message);
         }
+
+        // Function: SendMessage
+        // Used internally to send or request information from the Service via the <webSocket>. To
+        // be given a pre-made _message and _requestID. Provides an asynchronous <ConfigStateResponse>
+        // via the _callback parameter.
+        internal void SendMessage(string _message, string _requestID, Action<ConfigStateResponse> _callback)
```
NatGreenwood-UL: Could we name this (and the original) differently to distinguish the messages being sent?

ScreenControl_Unity/Assets/ScreenControl/Service/WebSocketReceiver.cs
```diff
@@ -14,8 +14,52 @@ namespace Ultraleap.ScreenControl.Service
...
+            string content;
+            if (configRequestQueue.TryPeek(out content))
+            {
+                // Parse newly received messages
+                configRequestQueue.TryDequeue(out content);
+                HandleConfigRequest(content);
+            }
+        }
+
+        void HandleConfigRequest(string _content)
```
NatGreenwood-UL: Can we rework this name to distinguish this (reqesting config _state_) from requesting a Config _change_? It threw me for a bit until I realised which this was.

ScreenControl_Web/src/Connection/ScreenControlServiceTypes.ts
```diff
@@ -53,6 +53,37 @@ export class ConfigState
     }
 }
 
+// class: ConfigRequest
+// Used to request the current state of the configuration on the Service. This is received as
+// a <ConfigState> which should be linked to a <ConfigStateCallback> via requestID to make
+// use of the data received.
+export class ConfigRequest
```
NatGreenwood-UL: Same deal here as in WebsocketReciever.cs: can we distinguish the config _change_ request and the config _state_ requests?

ScreenControl_Web/src/Connection/MessageReceiver.ts
```diff
@@ -100,7 +111,37 @@ export class MessageReceiver {
...
                     delete this.responseCallbacks[key];
-                    break;
+                    return;
+                }
+            };
+        }
+
+        console.log("Received a WebSocketResponse that did not match a callback." +
+            "This is the content of the response: \n Response ID: " + _response.requestID +
+            "\n Original request - " + _response.originalRequest +
```
NatGreenwood-UL: Could consider rearranging this so the original request is the _last_ thing printed here because it's likely to be multiple lines long and interrupt the readability? Would need to do the same in the Unity client.

NatGreenwood-UL (changes requested):
Few places where I think we need to distinguish the Config-related method names

ScreenControl_Unity/Assets/ScreenControl/Client/Scripts/Connection/ServiceConnection.cs
```diff
@@ -146,38 +146,26 @@ internal void SendMessage(string _message, string _requestID, Action<WebSocketRe
             webSocket.Send(_message);
         }
 
-        // Function: SendMessage
-        // Used internally to send or request information from the Service via the <webSocket>. To
-        // be given a pre-made _message and _requestID. Provides an asynchronous <ConfigStateResponse>
-        // via the _callback parameter.
-        internal void SendMessage(string _message, string _requestID, Action<ConfigStateResponse> _callback)
```
NatGreenwood-UL: Have we fully lost SendMessage now, in favour of RequestConfigState and SendChangeRequest?

MattGrayUL: I thought I did this and I couldn't find where! Will fix!

ScreenControl_Unity/Assets/ScreenControl/Client/Scripts/Connection/ServiceConnection.cs
```diff
@@ -145,5 +145,27 @@ internal void SendMessage(string _message, string _requestID, Action<WebSocketRe
 
             webSocket.Send(_message);
         }
+
+        // Function: RequestConfigState
+        // Used to request a <ConfigState> from the Service via the <webSocket>.
+        // Provides an asynchronous <ConfigState> via the _callback parameter.
+        public void RequestConfigState(Action<ConfigState> _callback)
```
NatGreenwood-UL: I just realised that this might be better off on the Configuration Manager as that's where one would request config _changes_, so it would make sense to live in the same place, for both Clients. 🤔

ScreenControl_Web/src/Connection/ServiceConnection.ts
```diff
@@ -6,11 +6,15 @@ import {
...
     CommunicationWrapper,
+    ConfigChangeRequest,
+    ConfigState,
+    ConfigStateCallback,
     ResponseCallback,
     WebSocketResponse
 } from './ScreenControlServiceTypes';
 import { ConnectionManager } from './ConnectionManager';
 import { v4 as uuidgen } from 'uuid';
+import { Guid } from 'guid-typescript';
```
NatGreenwood-UL: Do we need this? The Guid used below is generated by the uuidgen above, and I can't see a use of this anywhere?

NatGreenwood-UL (commented):
Just one comment, have tested both clients locally and get happy-looking results!

NatGreenwood-UL (approved):

PR #63, opened 2021-04-23
BL-696 Hand Presence Events on Connect

NatGreenwood-UL:


ScreenControl_Unity/Assets/ScreenControl/Service/Scripts/ServiceConnection/ClientConnectionManager.cs
```diff
@@ -38,27 +40,38 @@ private void Awake()
...
-                    HandPresenceEvent handsLostEvent = new HandPresenceEvent(HandPresenceState.HANDS_LOST);
-
                     _connection.SendHandPresenceEvent(handsLostEvent);
                 }
             }
+
+            // Cache handPresenceEvent when no clients are connected
+            if (activeConnections.Count == 0) {
+                missedHandPresenceEvent = handsLostEvent;
+            }
```
MattGrayUL: Should we not always cache the last event sent?

That way if a 2nd client joins, they will know the current state?

NatGreenwood-UL: Mmm that's a good point, though again that's something that starts to support multiple concurrent Clients, which is a use case we've been avoiding explicitly supporting until it can be considered on the whole. 🤔 I think we should leave it as-is for now, but we might want to put together a Confluence page of notes on this topic so that if that work does ever get agreed to that we can remember all these tidbits

MattGrayUL (approved):

PR #65, opened 2021-04-27
BL 734 prevent multiple tray icons

MattGrayUL:


ScreenControl_Service_Utilities/SC_ServiceUITray/SC_ServiceUITray/Program.cs
```diff
@@ -5,14 +5,27 @@
...
         static void Main()
         {
+            const string appName = "ScreenControl Tray";
+            bool createdNew;
+
+            mutex = new Mutex(true, appName, out createdNew);
+
+            if(!createdNew)
+            {
+                return;
```
NatGreenwood-UL, line 26: Would it be possible to find the app and give it focus instead of doing nothing? Gives users the flow of getting the screen as & when they hit the button regardless of having it up or not. Otherwise it may seem broken until they themselves realise there's a version running

MattGrayUL: This is as a result of launching a new Tray Icon, so I am unsure if it would be expected behavior for the ServiceUI to launch if it wasn't launched before (as the .exe they would open _should_ be named appropriately for the tray icon)

NatGreenwood-UL (approved):

PR #69, opened 2021-05-04
BL-558 Plugin System for Unity Client

MattGrayUL:
Added base plugin system (InputActionManager) and base plugin class to Unity Client. InputActions are now fed through this InputActionManager before being distributed

### DIFF
--- a/ScreenControl_Unity/Assets/ScreenControl/Client/Scripts/Connection/ConnectionManager.cs
+++ b/ScreenControl_Unity/Assets/ScreenControl/Client/Scripts/Connection/ConnectionManager.cs
@@ -36,15 +36,6 @@ namespace Ultraleap.ScreenControl.Client.Connection
         // A reference to the receiver that handles destribution of data received via the <currentServiceConnection> if connected.
         public static MessageReceiver messageReceiver;
 
-        // Delegate: ClientInputActionEvent
-        // An Action to distribute a <ClientInputAction> via the <TransmitInputAction> event listener.
-        public delegate void ClientInputActionEvent(ClientInputAction _inputData);
-
-        // Variable: TransmitInputAction
-        // An event for transmitting <ClientInputActions> that are received via the <messageReceiver> to
-        // be listened to.
-        public static event ClientInputActionEvent TransmitInputAction;
-
         // Variable: HandFound
         // An event allowing users to react to a hand being found when none has been present for a moment.
         public static event Action HandFound;
@@ -67,11 +58,11 @@ namespace Ultraleap.ScreenControl.Client.Connection
         // Group: Functions
 
         // Function: HandleInputAction
-        // Called by the <messageReceiver> to relay a <ClientInputAction> that has been received to any
-        // listeners of <TransmitInputAction>.
+        // Called by the <messageReceiver> to relay a <ClientInputAction> that has been received
+        // to the <InputActionManager> for distribution.
         public static void HandleInputAction(ClientInputAction _action)
         {
-            TransmitInputAction?.Invoke(_action);
+            InputActionManager.Instance.SendInputAction(_action);
         }
 
         // Function: AddConnectionListener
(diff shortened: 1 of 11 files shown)


ScreenControl_Unity/Assets/ScreenControl/Client/Scripts/Connection/ConnectionManager.cs
```diff
@@ -67,11 +58,11 @@ public static ServiceConnection serviceConnection
         // Group: Functions
 
         // Function: HandleInputAction
-        // Called by the <messageReceiver> to relay a <ClientInputAction> that has been received to any
-        // listeners of <TransmitInputAction>.
+        // Called by the <messageReceiver> to relay a <ClientInputAction> that has been received
+        // to the <InputActionManager> for distribution.
```
NatGreenwood-UL: Should we just move this function onto the `InputActionManager` rather than having this be the middleman when it doesn't need to?

ScreenControl_Unity/Assets/ScreenControl/Client/Scripts/Plugin System/InputActionPlugin.cs
```diff
@@ -0,0 +1,36 @@
...
+    // Class: InputActionPlugin
+    // A base class for Plugins which are used by the <InputActionManager> to manipulate the
+    // <ClientInputAction> data received from the Service.
+    public class InputActionPlugin : MonoBehaviour
+    {
+        // Variable: TransmitInputAction
+        // An event for transmitting <ClientInputActions> as they pass through this plugin.
+        // This can be used to access the data as it is used by a specific plugin, as to intercept
+        // the full cycle of plugins that the <InputActionManager> references.
+        public event InputActionManager.ClientInputActionEvent TransmitInputAction;
```
NatGreenwood-UL: Do we need to _require_ this event on every plugin? We're using `RunPlugin` directly from the `InputActionManager`, so we don't directly _need_ it. Though I can see us listening directly to events from specific plugins I'm not sure it's needed in the base.

ScreenControl_Unity/Assets/ScreenControl/Client/Scripts/Plugin System/InputActionManager.cs
```diff
@@ -0,0 +1,75 @@
...
+        // <plugins>
+        public static event ClientInputActionEvent TransmitRawInputAction;
+
+        public static InputActionManager Instance;
+
+        // Variable: plugins
+        // A pre-defined plugin array of <ToggleablePlugins> that modify incoming <ClientInputActions>
+        // based on custom rules.
+        [Tooltip("These plugins modify InputActions and are performed in order.")]
+        [SerializeField] ToggleablePlugin[] plugins;
```
NatGreenwood-UL, line 35: Is this array re-arrangable in the inspector UI? Not huge if not, just would be nicer UX.

MattGrayUL: It will be in 2020+ (or 2021 not sure, Unity added rearranging Inspector arrays recently)

ScreenControl_Unity/Assets/ScreenControl/Client/Scripts/Plugin System/InputActionPlugin.cs
MattGrayUL: We don't _need_ to, but my thinking is that then any plugin has an event that can be listened to and the standard behavior should be to call the base to trigger that.

I think it becomes hidden as a concept if we don't include it here and if nothing listens it comes at very little (or no?) cost

ScreenControl_Unity/Assets/ScreenControl/Client/Scripts/Connection/ConnectionManager.cs
MattGrayUL: I agree here, we should just send the inputactions to the inputactionmanager from the messagereceiver directly

NatGreenwood-UL (approved):

PR #70, opened 2021-05-06
Feature/bl 739 touch plane push

MattGrayUL:


ScreenControl_Unity/Assets/ScreenControl/Service/Scripts/Interactions/InteractionModules/InteractionModule.cs
```diff
@@ -59,9 +59,9 @@ void Update()
         // This is the main update loop of the interaction module
         protected virtual void UpdateData(Leap.Hand hand) { }
 
-        protected void SendInputAction(ScreenControlTypes.InputType _inputType, Positions _positions, float _distanceFromScreen, float _progressToClick)
+        protected void SendInputAction(ScreenControlTypes.InputType _inputType, Positions _positions, float _progressToClick)
```
NatGreenwood-UL, line 62: How come we're stripping the distance out of this? Do we not need to keep it for the plugin to filter the cursor to limited distances?

ScreenControl_Web/src/Configuration/ConfigurationTypes.ts
```diff
@@ -57,6 +58,27 @@ export interface HoverAndHoldInteractionSettings {
...
+// using the <ConfigurationManager>.
+//
+// Like all of the Settings classes found in this file, all members are optional. If you do
+// not modify a member of this class, its value will not change when the instance is sent to
+// ScreenControl Service.
+export interface TouchPlaneInteractionSettings {
+    // Property: TouchPlaneDistanceCM
+    // This determines how far (in cm) the TouchPlane is from the screen surface. This
+    // represents the plane that the user must pass to begin and end a click event.
+    TouchPlaneDistanceCM: number;
```
NatGreenwood-UL: Can we differentiate this and the below further? "TouchPlaneActivationDistance" vs "TouchPlaneStartDistance" or something like that? Having one that is the same as the other but with one more word is a little confusing

NatGreenwood-UL (commented):
Couple minor comments

ScreenControl_Unity/Assets/ScreenControl/Service/Scripts/Interactions/InteractionModules/InteractionModule.cs
MattGrayUL: Distance actually exists in Positions, so we were sending it twice and ignoring one. When the WS version is created, it sends the distance too :)

ScreenControl_Web/src/Configuration/ConfigurationTypes.ts
MattGrayUL: TouchPlaneStartDistanceCM is only intended to be around until we user test and find a valid distance, this isn't a value we should keep in the UI (or in the config files) as it is hard to explain to users and is unlikely to be wanted to change.

I added it on all sides of the WS just in case we needed it while testing but fully intend to remove it (so isn't overly worth worrying about)

ScreenControl_Unity/Assets/ScreenControl/Service/Scripts/Interactions/InteractionModules/InteractionModule.cs
NatGreenwood-UL: Ahhhh ok I didn't realise it was inside the `positions`, all cool then

ScreenControl_Web/src/Configuration/ConfigurationTypes.ts
NatGreenwood-UL: That's true but until we do this I'd rather it be less confusing internally 😅

NatGreenwood-UL (approved):

PR #71, opened 2021-05-07
BL-728 TouchFree in installer

NatGreenwood-UL:
* Installer now (optionally) installs TouchFree as well! (defaults to installing, can be disabled)
* Generated a new AppID for this installer so it doesn't collide with previous SC Service installations. 
  * Please uninstall SC_Service if you had it installed previously! They won't collide but it saves you from accidentally using the wrong setup.
* If you _do_ install TouchFree, the option at the end to launch the Service UI will be _replaced_ with the option to launch TouchFree instead (as you will be able to open the ServiceUI from TouchFree's once it has a UI again).


MattGrayUL (approved):
Tested the installer, all looks good to me for the brief

PR #74, opened 2021-05-12
BL-749 Interaction Settings UI squish fix

MattGrayUL:
Removed SettingsUIScrollElementScaler and replaced with proper LayoutGroup settings


NatGreenwood-UL (approved):
Tried this out in 16:9, 9:16 and 4:3 and they all look good, not entirely clear on what you changed here to make this happen so would appreciate a quick rundown before we merge both this and the TouchFree version. 👍

PR #86, opened 2021-05-19
Updated API versions on both Clients, Service and CI

MattGrayUL:


NatGreenwood-UL (approved):
Tested this with a Service UI build from this branch and an outdated service, connection is refused, as expected. We should consider adding some explicit messaging to the Service UI and TouchFree itself when this happens though, though I guess as all three apps are installed by the same installer, that's not likely to come up with users 🤔

PR #93, opened 2021-05-27
Bug/bl 770 bl 771 touch free UI opening at wrong times

MattGrayUL:


NatGreenwood-UL (approved):
Tested locally, couldn't reproduce either bug any more 🎉

PR #97, opened 2021-06-04
BL-773 Point the file explorer to the current directory rather than the current file path

MattGrayUL:


NatGreenwood-UL (approved):
Tested this locally, looks like we're getting our intended functionality now 👌

PR #100, opened 2021-06-14
TF-74 - Change the regex used to split the incoming JSON to handle whitespace

NatGreenwood-UL:


MattGrayUL (approved):
Have tested this locally and it works! I approve

PR #101, opened 2021-06-14
Touch free cursor size change

NatGreenwood-UL:


TF_Application/Assets/TouchFree/Tooling/Scripts/Cursors/DotCursor.cs
```diff
@@ -364,6 +362,8 @@ protected virtual void SetCursorLocalScale(float _scale)
         {
             cursorLocalScale = new Vector3(_scale, _scale, _scale);
             cursorBorder.transform.localScale = cursorLocalScale;
+
+            maxRingScale = cursorMaxRingSize;
```
MattGrayUL: the use of maxRingScale can be removed and replaced with directly using cursorMaxRingScale

TF_Application/Assets/TouchFree_Application/Scenes/TouchlessOverlay.unity
```diff
@@ -233,6 +233,16 @@ PrefabInstance:
   m_Modification:
```
MattGrayUL, line 233: All of these changes should come from the prefab and not from the TouchlessOverlay scene

MattGrayUL (commented):
These changes work great. They need to be replicated in the Tooling directly and then the tooling imported into TouchFree App.|

- The fade of the cursor should probably be slower, more like 15.
- The curve of the squeeze feels good to me
- Values should be applied to the prefab (and to the defaults in the class too)
- I find the ring thickness is too thick, and the fill cursor was made to closely match it, so we should ensure they match either way
- Ensure tooling documentation for cursors are up to date

TF_Service_and_Tooling_Unity/Assets/TouchFree/Tooling/Prefabs/Cursors/RingCursor.prefab
```diff
@@ -300,7 +300,7 @@ GameObject:
   m_Icon: {fileID: 0}
   m_NavMeshLayer: 0
   m_StaticEditorFlags: 0
-  m_IsActive: 1
+  m_IsActive: 0
```
MattGrayUL: I think we want this to be on by default (and overridden in the TouchFreeApp Scene)

MattGrayUL (approved):
All seems good

PR #102, opened 2021-06-14
Bug/bl 774 cti appears when hands present

MattGrayUL:


NatGreenwood-UL (commented):
All looks good, reproduced the bug on develop, then couldn't on this branch. 👍

NatGreenwood-UL (approved):
Actually approving, forgot default state was Commenting 😅

PR #103, opened 2021-06-16
Created InputActionManager system in web to match Unity Tooling

MattGrayUL:


TF_Tooling_Web/src/Plugins/InputActionManager.ts
```diff
@@ -0,0 +1,73 @@
...
+// via the  <TransmitInputAction> event which should be listened to by any class hoping to make
+// use of incoming <TouchFreeInputActions>.
+export class InputActionManager extends EventTarget {
+    // Event: TransmitInputAction
+    // An event for transmitting <TouchFreeInputActions> that are received via the <messageReceiver> to
+    // be listened to.
+
+    // Event: TransmitInputActionRaw
+    // An event for immediately transmitting <TouchFreeInputActions> that are received via the
+    // <messageReceiver>to be listened to.
```
NatGreenwood-UL: I'd explicitly call out that this is _before_ plugins are executed as opposed to the other which is post-plugins. Also there's a missing space here between the close angle-brace and the word which may mess up NaturalDocs' generation

TF_Tooling_Web/src/Plugins/InputActionManager.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    // Variable: instance
+    // The instance of the singleton for referencing the events transmitted
+    static _instance: InputActionManager;
+
+    static plugins: Array<InputActionPlugin>;
+
+    public static get instance() {
+        if (InputActionManager._instance == null) {
```
NatGreenwood-UL: I'm not entirely sure what will happen with this in JavaScript, I think that uninitialised objects will compare equal to null, but the strictly "safer" way to do this would be using `===` and in that case we might need to set it to null up on line 20.

TF_Tooling_Web/src/Plugins/InputActionManager.ts
```diff
@@ -0,0 +1,73 @@
...
+        let action = _action;
+
+        if (this.plugins != null) {
+            for (var i = 0; i < this.plugins.length; i++) {
+                let modifiedAction = this.plugins[i].RunPlugin(action);
+
+                if (modifiedAction != null) {
+                    action = modifiedAction;
+                }
+                else {
```
NatGreenwood-UL: Bit of a mishmash of brace styles here; can we stick to one-line (`} else {`) or line-per-brace?

TF_Tooling_Web/src/Plugins/InputActionPlugin.ts
```diff
@@ -0,0 +1,40 @@
+import { TouchFreeInputAction } from "../TouchFreeToolingTypes";
+
+export class InputActionPlugin extends EventTarget{
```
NatGreenwood-UL: I think we should make [this class `abstract`](https://www.typescriptlang.org/docs/handbook/2/classes.html#abstract-classes-and-members) as it is in the Unity version to prevent people trying to _use_ this plugin

NatGreenwood-UL (commented):
Couple minor TypeScript oddities to change just to be as safe as we can be, and a comment that needs a touch of tweaking

NatGreenwood-UL (approved):
All looks good on this end, will throw a ticket in the BL to discuss the hide/show cursor issues we've noticed 👍

PR #104, opened 2021-06-22
Feature/bl 778 single click taskbar to open UI

MattGrayUL:


NatGreenwood-UL (approved):
All looks good, the second iteration of the minimising works much faster so I think we should stick with that 👍

PR #105, opened 2021-06-22
Bug/tf 78 various UI and flow bugs

MattGrayUL:


NatGreenwood-UL (approved):
Manually tested that all elements are fixed, they are, but the install directory is persisted in upgrade scenarios so I had to uninstall/reinstall to test the change to Program Files. Worth noting in release notes.

PR #109, opened 2021-06-23
Feature/bl 781 remove need for admin login

MattGrayUL:


NatGreenwood-UL (approved):
Got this tested and can confirm that Windows doesn't get in the way any further, though reproducing it was a little fiddly.

To reproduce in future; 
* Open User Account Settings in windows
* Set the setting to the highest security option
* *Restart the computer*
* Launch the Service UI from the tray icon
* You do _not_ get the warning from Windows

PR #111, opened 2021-06-23
Feature/bl 788 track nearest bone

MattGrayUL:


NatGreenwood-UL (commented):
Looks good overall, want to check with you in the morning about the change to AirPush using the fingertip because of the whole finger-curl issue. Would be nice to try this out solidly on kiosks before comitting to that change.

MattGrayUL:
> Looks good overall, want to check with you in the morning about the change to AirPush using the fingertip because of the whole finger-curl issue. Would be nice to try this out solidly on kiosks before comitting to that change.

This was a mistake, I will change to the previously accepted tracking position

MattGrayUL:
> Looks good overall, want to check with you in the morning about the change to AirPush using the fingertip because of the whole finger-curl issue. Would be nice to try this out solidly on kiosks before comitting to that change.

It uses the 'index stable' position which is the previously accepted position (near the proximal joint)

NatGreenwood-UL (approved):
Looked back over this, all looks good

PR #112, opened 2021-06-23
Feature/bl 783 update icons etc

MattGrayUL:


TF_Service_Utilities/Installer/TouchFree_Installer.iss
```diff
@@ -4,7 +4,7 @@
 #define CompanyURL "https://ultraleap.com"
 #define ProductName "TouchFree"
 #define Publisher "Ultraleap Inc."
-#define ReleaseVersion "1.0.0-alpha5"
+#define ReleaseVersion "2.0.0-beta1"
```
NatGreenwood-UL: We could consider replacing this via the same method we replace the entries in the version.txt so that we don't have to replace them each time we update

NatGreenwood-UL (approved):
Looks good overall, had to restart to get the taskbar icons to update. One comment about maybe streamlining the version iteration process, doesn't _need_ to be resolved here but could be whilst we're looking

PR #114, opened 2021-06-24
Feature/bl 789 cursor based UI adjustments

MattGrayUL:


TF_Application/Assets/Examples/FillCursor/Scripts/FillCursor.cs
```diff
@@ -24,6 +24,14 @@ public class FillCursor : TouchlessCursor
 
     public Image fillRingImage;
 
+    // Variable: minRingThickness
+    // The minimum thickness the ring can be.
+    public float minRingThickness = 1.05f;
```
NatGreenwood-UL: Should these maybe be public but readonly? Is there a situation where we want people to be able to edit this live?

TF_Application/Assets/TouchFree/Tooling/Scripts/Cursors/DotCursor.cs
```diff
@@ -60,10 +61,13 @@ public class DotCursor : TouchlessCursor
         // transparent. It has to be scaled to match the ring itself.
         public RectTransform ringMask;
 
-        // Variable: ringThickness
-        // Used to set the thickness of the ring itself (i.e. the distance between the inner and
-        // outer edges of the ring)
-        public float ringThickness = 1.5f;
+        // Variable: minRingThickness
+        // The minimum thickness the ring can be.
+        public float minRingThickness = 1.5f;
```
NatGreenwood-UL: Should this and the below not be overrides of the baseclass?

TF_Application/Assets/Examples/FillCursor/Scripts/FillCursor.cs
NatGreenwood-UL: Also much like the comment on the DotCursor, these two should maybe be overrides just to save any minor access issues

TF_Application/Assets/TouchFree/Tooling/Scripts/Cursors/DotCursor.cs
MattGrayUL: I was thinking that, but i'm not enjoying the dependency the base has on rings although because of the thickness value we currently have, there is no getting around it until we have per-cursor configs. Will move to the base for now though

TF_Application/Assets/Examples/FillCursor/Scripts/FillCursor.cs
MattGrayUL: > Should these maybe be public but readonly? Is there a situation where we want people to be able to edit this live?

For Tooling users, I think these values shouldn't be readonly. Some may want very thick or very thin cursor rings, the min/max defaults here are tailored towards TouchFree App

NatGreenwood-UL: Yeah but you can override them in an inheriting class, which lets you set your own values. The readonly thing more affects whether you can tweak it live. I guess because you can't loosen access modifiers when overriding though this might be better to leave open so we don't restrict customer cursors.

TF_Application/Assets/TouchFree/Tooling/Scripts/Cursors/DotCursor.cs
NatGreenwood-UL: Oh I thought they were already _on_ the baseclass and was just commenting that they should be overrides instead of declaring a second member with the same name (making the base class' inaccessible). I think for now, given that all of our cursors use rings, it's fine to have rings on the base class, though we may want to switch that in future. We could even consider making a "RingCursor" _interface_ so we don't have to worry about a middle base-class, if there is no ring-specific but for-every-ring functionality.

NatGreenwood-UL (approved):
All looks good with the toggles now they're in!

PR #117, opened 2021-06-30
Feature/bl 790 offer touch plane trackng bone option in UI

MattGrayUL:


NatGreenwood-UL (approved):
All looks good with the PR, have tried it out and it seems great, with our new process you should probably get Dean to have a quick check over this if we think it's important for him.

One though did emerge reviewing it, which was the Enum name change from ALL_CAPS_UNDERSCORED. Obviously it now matches the rest of the file it's now in, but the caps is good practice for constants. That said, enums aren't _quite_ constants? Not one for right now, but worth thinking about when we do our Development Practices meeting.

PR #118, opened 2021-07-01
Feature/tf 82 remove smoothing

MattGrayUL:


NatGreenwood-UL (approved):
Code changes look good, big fan of how things feel with _just_ the Deadzone now 👍

PR #123, opened 2021-08-02
[TF-119] Blocked interactions when users hands initialize within the TouchPlane zone

MattGrayUL:
- [x] Code Review

### DIFF
--- a/TF_Service_and_Tooling_Unity/Assets/TouchFree/Service/Scripts/Interactions/InteractionModules/TouchPlanePushInteraction.cs
+++ b/TF_Service_and_Tooling_Unity/Assets/TouchFree/Service/Scripts/Interactions/InteractionModules/TouchPlanePushInteraction.cs
@@ -20,7 +20,7 @@ namespace Ultraleap.TouchFree.Service
 
         // Used to ignore hands that initialise while past the touchPlane.
         // Particularly for those that are cancelled by InteractionZones
-        bool handWasCancelled = true;
+        bool handReady = false;
 
         [Header("Dragging")]
         public float dragStartDistanceThresholdM = 0.01f;
@@ -39,7 +39,7 @@ namespace Ultraleap.TouchFree.Service
                 pressComplete = false;
                 isDragging = false;
                 pressing = false;
-                handWasCancelled = true;
+                handReady = false;
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


TF_Service_and_Tooling_Unity/Assets/TouchFree/Service/Scripts/Interactions/InteractionModules/TouchPlanePushInteraction.cs
```diff
@@ -18,6 +18,10 @@ public class TouchPlanePushInteraction : InteractionModule
 
         private Vector2 downPos;
 
+        // Used to ignore hands that initialise while past the touchPlane.
+        // Particularly for those that are cancelled by InteractionZones
+        bool handWasCancelled = true;
```
PeterNancollis-UH: Might not need changing, but is it only the case that this is true when a hand is cancelled? 
From looking through the code it appears this would also be true when a hand first appears, regardless of whether the interaction zone has cancelled it or if it has just been detected for the first time. I might be missing something though.

MattGrayUL: Yes so this also is set to true every time tracking is lost (and is true by default for when you first track a hand), so it may be that we want to rename the variable. I couldn't think of anything more suitable but if you can, we can change it :) 

Ultimately it should be true whenever we don't expect to see a hand and remain true until the hand is detected _and_ further from the screen than the touchplane

PeterNancollis-UH: `handWasNull` would be the simplest option as it says exactly what we know about the hand, but it doesn't capture the part about it only clearing when your hand is outside the touch plane.
`handNotReady` helps imply that there is a condition attached, but is confusing to be testing if `!handNotReady` as it's a double negative. 
`handReady` would read better, but would mean inverting some of your logic for it to make sense. ie `if (handReady)` in place of `if (!handWasCancelled)` and making it be false when the hand is null

MattGrayUL: I will implement handReady as that makes the most sense to me too :)

PeterNancollis-UH (approved):

PR #125, opened 2021-08-02
[TF-111] Added cursor settings to 'ToHide' array

MattGrayUL:


PeterNancollis-UH (approved):
Checked setting applied in the prefab as described and that cursor thickness setting is hidden when cursor disabled in the application.

PR #126, opened 2021-08-02
[TF-90] Automatic Cursor Window Resizing

MattGrayUL:
Added a scaling remap to TouchFreeMain to adjust the cursor window size based on the cursor size.

Window sizes allow for max thickness rings at all times

- [x] Code Review
- [x] Test plan updated


MattGrayUL:
- [ ] Code Review
- [ ] Test plan updated

PeterNancollis-UH (commented):
I noticed this has fallen behind develop by some way while sitting in review. There's no apparent conflicts but would we usually update the branch so we don't get any surprises during testing?

MattGrayUL:
> I noticed this has fallen behind develop by some way while sitting in review. There's no apparent conflicts but would we usually update the branch so we don't get any surprises during testing?

I think that's worth doing for any that we currently have open yeah

PeterNancollis-UH (approved):

PR #127, opened 2021-08-02
Bug/tf 120 tooling config change errors

MattGrayUL:
- [x] Code Review

### DIFF
--- a/TF_Service_and_Tooling_Unity/Assets/TouchFree/Tooling/Scripts/Configuration/ConfigurationManager.cs
+++ b/TF_Service_and_Tooling_Unity/Assets/TouchFree/Tooling/Scripts/Configuration/ConfigurationManager.cs
@@ -36,20 +36,20 @@ namespace Ultraleap.TouchFree.Tooling.Configuration
             string jsonContent = "";
             jsonContent += "{\"action\":\"";
             jsonContent += action + "\",\"content\":{\"requestID\":\"";
-            jsonContent += requestID + "\",";
+            jsonContent += requestID + "\"";
 
-            if(_interaction != null)
+            if (_interaction != null)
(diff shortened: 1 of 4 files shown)


TF_Service_and_Tooling_Unity/Assets/TouchFree/Tooling/Scripts/Configuration/ConfigurationManager.cs
```diff
@@ -36,20 +36,20 @@ public static void RequestConfigChange(InteractionConfig _interaction, PhysicalC
             string jsonContent = "";
             jsonContent += "{\"action\":\"";
             jsonContent += action + "\",\"content\":{\"requestID\":\"";
-            jsonContent += requestID + "\",";
+            jsonContent += requestID + "\"";
 
-            if(_interaction != null)
+            if (_interaction != null)
             {
+                jsonContent += ",";
```
PeterNancollis-UH: While testing in an app of my own I found if we:

- request the config, 
- store a copy of that locally, 
- modify only one of the interaction or physical configs 
- request a change of the config, with both my modified and unmodified configs
- then the service doesn't respond or make the requested change

After some discussion we found that this comma was to blame. When loading the config received from the server the values are applied without calling the setters so the list of config options is empty. This meant this comma gets added, even though the json blob is returned empty and the service throws and error.
This is also true for the physical config below.

PeterNancollis-UH: Looks good based on the commit in response to this and works as expected in my test app

PeterNancollis-UH (approved):

PR #130, opened 2021-08-09
Feature/tf 123 utility apps 64bit

MattGrayUL:
- [x] Code Review


PeterNancollis-UH (approved):
Was pleased to see when testing that memory usage didn't increase noticeably with the switch to 64bit. There is slightly higher usage for the service wrapper, but lower usage for the tray icon so it balanced out.

PR #131, opened 2021-08-09
Bug/tf 126 tooling up event position correction

MattGrayUL:
- [x] Code Review


TF_Application/Assets/TouchFree/Tooling/Scripts/Connection/MessageReceiver.cs
```diff
@@ -132,14 +132,19 @@ void HandleConfigState(ConfigState _configState)
             }
         }
 
+        Vector2 lastKnownCursorPosition = new Vector2();
```
PeterNancollis-UH: This should probably be declared at the top with other properties. Also lacks documenting comments

TF_Service_and_Tooling_Unity/Assets/TouchFree/Tooling/Scripts/Connection/MessageReceiver.cs
```diff
@@ -132,14 +132,19 @@ void HandleConfigState(ConfigState _configState)
             }
         }
 
+        Vector2 lastKnownCursorPosition = new Vector2();
```
PeterNancollis-UH: Reminder to make sure changes made to the copy of this file in TF_Application are also made to the tooling

TF_Tooling_Web/src/Connection/MessageReceiver.ts
```diff
@@ -152,16 +152,20 @@ export class MessageReceiver {
...
     // <actionCullToCount>. If any remain, sends the oldest <TouchFreeInputAction> to
     // <InputActionManager> to handle the action.
+    // UP <InputType>s have their positions set to the last known position to ensure
+    // input events trigger correctly.
     CheckForAction(): void {
         while (this.actionQueue.length > this.actionCullToCount) {
             if (this.actionQueue[0] !== undefined) {
                 // Stop shrinking the queue if we have a 'key' input event
-                if (this.actionQueue[0].InteractionFlags & InputType.MOVE) {
+                if ((this.actionQueue[0].InteractionFlags & InputType.MOVE) != 0) {
```
PeterNancollis-UH: This change seems unrelated to the cursor position. What is it for?

PeterNancollis-UH (changes requested):

TF_Tooling_Web/src/Connection/MessageReceiver.ts
MattGrayUL: That's my bad, I thought the bit comparison was acting oddly so I added an explicit check but it wasn't and forgot to remove. Will revert

TF_Application/Assets/TouchFree/Tooling/Scripts/Connection/MessageReceiver.cs
MattGrayUL: This won't include documenting comments due to it being private but I will move it to the top and add some comments in-file to explain its use (in TF App, the Unity Tooling itself and the Web Tooling)

PeterNancollis-UH (changes requested):
The change to the web tooling appears to be preventing UP pointer events from triggering so if I modify the quick-start to activate the buttons on UP they never activate.

PeterNancollis-UH (approved):

PR #132, opened 2021-08-10
Feature/tf 89 automatically capture resolution during setup

MattGrayUL:
- [x] Code Review


TF_Service_and_Tooling_Unity/Assets/TouchFree/ServiceUI/Scripts/Configuration/Screens/QuickSetupCompleteScreen.cs
```diff
@@ -1,4 +1,5 @@
 ﻿using UnityEngine;
+using Ultraleap.TouchFree.ServiceShared;
```
PeterNancollis-UH: This doesn't appear to be used, is it necessary?

TF_Service_and_Tooling_Unity/Assets/TouchFree/ServiceUI/Scripts/Configuration/ConfigUI/ManualConfigUI.cs
```diff
@@ -43,6 +43,12 @@ private void Awake()
             resetToDefaultWarning.SetActive(false);
         }
 
+        protected override void OnEnable()
+        {
+            base.OnEnable();
+            VirtualScreen.CaptureCurrentResolution();
```
PeterNancollis-UH, line 49: In testing this I found that the captured resolution is applied to the visuals of the manual setup screen, but it isn't applied to the cursor position until you leave the manual setup screen. Is this the expected behaviour?

I found myself a little confused as the cursor was wrong while on the screen, then it was magically correct after I left.

TF_Service_and_Tooling_Unity/Assets/TouchFree/ServiceUI/Scripts/Configuration/Screens/QuickSetupCompleteScreen.cs
MattGrayUL: It isn't, I think this is carry-over from one of the other commits. Will delete

TF_Service_and_Tooling_Unity/Assets/TouchFree/ServiceUI/Scripts/Configuration/ConfigUI/ManualConfigUI.cs
MattGrayUL: This is interesting as I would not expect that to be the case. It changes the text file which the Service listens to and should immediately update the Interaction (and so, the cursor position)

PeterNancollis-UH: The latest commit fixes this issue. The cursor now updates to the newly captured resolution when the screen loads

PeterNancollis-UH (approved):

PR #134, opened 2021-08-13
TF-134 Optimize snapping plugin for web

MattGrayUL:
Changed elementsFromPoint call to check for point within rect bounds instead

- [x] Code Review

### DIFF
--- a/TF_Tooling_Web/Plugins/SnappingPlugin/src/Ray.ts
+++ b/TF_Tooling_Web/Plugins/SnappingPlugin/src/Ray.ts
@@ -59,7 +59,7 @@ export class Ray {
         pos.y = pos.y + (vector.y / 2);
 
         while (distance > 1) {
-            hasSnap = Ray.Hit(pos, element);
+            hasSnap = Ray.Hit(pos, element.getBoundingClientRect());
 
             // If we changed state, we reverse the direction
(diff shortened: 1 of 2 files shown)


TF_Tooling_Web/Plugins/SnappingPlugin/src/Ray.ts
```diff
@@ -88,8 +88,11 @@ export class Ray {
         return pos;
     }
 
-    public static Hit(position: Vector2, element: Element): boolean {
-        let elementsAtPos: Element[] = document.elementsFromPoint(position.x, position.y);
-        return elementsAtPos.find((value: Element, index: number, obj: Element[]) => value === element) !== undefined;
+    public static Hit(position: Vector2, rect: DOMRect): boolean {
+
+        return !(rect.right < position.x ||
```
PeterNancollis-UH: Either left/right or top/bottom logic looks reversed though the result does appear to behave as expected.

Might be easier to read as using AND logic rather than NOT OR as it is now.

I expect a hit to be:
```
position.x > rect.left && 
position.x < rect.right && 
position.y > rect.bottom && 
position.y < rect.top
```

PeterNancollis-UH: The logic error was down to me not realising the y axis origin is at the top of the page.

PeterNancollis-UH (approved):

PR #135, opened 2021-08-18
Feature/tf 131 adjust default grab

MattGrayUL:
- [x] Code Review
- [x] Product Owner Review (Optional)


TF_Service_and_Tooling_Unity/Assets/TouchFree/Service/Scripts/Interactions/InteractionModules/GrabInteraction.cs
```diff
@@ -172,7 +171,7 @@ private void HandlePotentialUnclick()
             // Check if an unclick is needed, and perform if so
             if (pressing)
```
PeterNancollis-UH: I think we no longer need to check for pressing here as our only path to this function is when pressing is true

TF_Service_and_Tooling_Unity/Assets/TouchFree/Service/Scripts/Interactions/GeneralisedGrabDetector/GeneralisedGrabDetector.cs
```diff
@@ -7,197 +7,49 @@
...
-        }
-
-        private void ResolveClassicPinch(Leap.Hand hand)
-        {
-            float pinchStrength = hand.PinchStrength;
             if (grabbing)
             {
-                grabbing = (pinchStrength >= unpinchThreshold);
-                GeneralisedGrabStrength = ServiceUtility.MapRangeToRange(Mathf.Clamp(pinchStrength, 0, unpinchThreshold), 0, unpinchThreshold, 0, 1);
+                return true;
```
PeterNancollis-UH: We could tidy this up even further by just returning `grabbing`

PeterNancollis-UH (approved):

PR #140, opened 2021-08-19
TF-152 Fix Cursor window opacity issues

MattGrayUL:
changed to capture window ptr by name. Check the window is transparent while it is set to cursor mode

- [x] Code Review

### DIFF
--- a/TF_Application/Assets/TouchFree_Application/Scripts/TransparentWindow.cs
+++ b/TF_Application/Assets/TouchFree_Application/Scripts/TransparentWindow.cs
@@ -166,25 +166,26 @@ namespace Ultraleap.TouchFree
 #if !UNITY_EDITOR
 		    if (clickThroughEnabled)
 		    {
-                int xPos = Mathf.RoundToInt(position.x);
-                int yPos = Mathf.RoundToInt(position.y);
-
-                if (xPos + TouchFreeMain.CursorWindowSize > 0 && xPos < Display.main.systemWidth && yPos + TouchFreeMain.CursorWindowSize > 0 && yPos < Display.main.systemHeight)
-                {
-                    SetWindowPos(hwnd,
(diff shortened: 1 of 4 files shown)


TF_Application/Assets/TouchFree_Application/Scripts/TransparentWindow.cs
```diff
@@ -166,18 +166,26 @@ void Update()
 #if !UNITY_EDITOR
 		    if (clickThroughEnabled)
```
PeterNancollis-UH: The indentation is a bit off for this if statement (including it's closing brace)

PeterNancollis-UH:
This is almost there, the cursor now presents correctly even if I switch windows behind the CTI. 

Remaining issues:
- Sometimes the Settings UI opens when clearing the CTI at which point it appears in a maximised window rather than full screen
- Sometimes it can take a couple of repeated clicks to get the settings UI to open after the CTI clears

TF_Application/Assets/TouchFree_Application/Scripts/TransparentWindow.cs
MattGrayUL: image

Interestingly I don't have that issue in my IDE will try to address though

PeterNancollis-UH: I think it's because there's a mixture of tabs and spaces on those lines, so I guess github renders that differently to your IDE.

PeterNancollis-UH:
The latest commit seems to be a solid fix. Have tried to break it using all the methods we've found to cause the issue and in all cases it now behaves as we would expect with the CTI clearing normally and cursor returning to the transparent window state

PeterNancollis-UH (approved):

PR #141, opened 2021-08-24
BugFix: Cursor position gets inverted by Input System

PeterNancollis-UH:
The input system was directly modifying the incoming InputAction which could cause issues for plugins that also access the InputAction, especially if they store a reference.

- [x] Code Review
- [x] Developer Testing


NatGreenwood-UL (commented):
Code looks good, just gonna wait for a build to try out to confirm the issue is resolved 👍

NatGreenwood-UL (approved):
Resolved all the inversion issues I was having locally, looks great! 👍

PR #149, opened 2021-09-10
Feature/tf 169 combine overlay visuals into settings app

MattGrayUL:
- [x] Test plan updated (by PR creator) - Update will be implemented in epic merge
- [x] Code Review


PeterNancollis-UH (commented):
Everything looks good after review and basic testing. Some odd behaviour currently but we expect these items to be resolved by work on other tickets on this epic branch.

- Changing settings after accessing from tray icon doesn't affect TF App until it's restarted
- We can currently access the settings from two locations
- Opening TF app settings after making changes from tray icon results in those changes being lost

MattGrayUL:
> * Opening TF app settings after making changes from tray icon results in those changes being lost

That is not something I would expect... it will be fine after we complete the rest of the epic, but I still wouldn't expect TF app settings to save the values on open... I assume TFApp was running with cursor whilst you changed values in ServiceUI?

PR #151, opened 2021-09-13
TF-171 update the overlay visuals in realtime

MattGrayUL:
- [x] Code Review


PeterNancollis-UH:
Might not be directly related to this PR, but enabling the CTI in the new setting UI doesn't work I think because the default path to the CTI files is wrong. This means you have to change the file before you will see a CTI.

Showed up when testing if enabling the CTI would live update, which it does once the path is updated

TouchFreeServiceUI_3Oyqp53dv3

PeterNancollis-UH:
We have agreed that the above issue can be resolved separately so doesn't need to hold up this PR.

PeterNancollis-UH (approved):

PR #152, opened 2021-09-13
TF-172: Remove Settings screen from overlay app

NatGreenwood-UL:
- [x] Code Review


TF_Application/Assets/TouchFree_Application/Scripts/TransparentWindow.cs
```diff
@@ -199,30 +196,12 @@ public void SetPosition(Vector2 value)
             position.y = Display.main.systemHeight - position.y - (TouchFreeMain.CursorWindowSize / 2);
         }
 
-
-        bool ctiActive = false;
-        void CTIActivated()
```
PeterNancollis-UH, line 204: I think the code was a little more readable with this removed method where we linked the `OnCTIActive` event to the `CTIActivated` handler. It does make for very thin handlers, but I think it's more self explanatory

NatGreenwood-UL: Would your preference be the return of these wrappers then or simply to rename the Handlers we still have? I merged these together because the underlying "ClickThrough" events were used directly by the ScreenManager which no longer calls them, so as there is only one user of them, having two seperate functions for it felt like overkill.

PeterNancollis-UH: I think it would read fine if we just renamed the methods. The click through enable/disable is self explanatory inside that method anyway and feels like a part of the process of enabling the CTI.

PeterNancollis-UH (approved):

PR #153, opened 2021-09-14
Feature/tf 194 warn when insufficient permissions in UI

MattGrayUL:
- [x] Code Review

### DIFF
--- a/TF_Service_and_Tooling_Unity/Assets/TouchFree/ServiceShared/Scripts/Configuration/ConfigFiles/ConfigFile.cs
+++ b/TF_Service_and_Tooling_Unity/Assets/TouchFree/ServiceShared/Scripts/Configuration/ConfigFiles/ConfigFile.cs
@@ -81,7 +81,14 @@ namespace Ultraleap.TouchFree.ServiceShared
 
         protected void SaveConfig_Internal(TData config)
         {
-            File.WriteAllText(_ConfigFilePath, JsonUtility.ToJson(config, true));
+            try
+            {
+                File.WriteAllText(_ConfigFilePath, JsonUtility.ToJson(config, true));
+            }
+            catch
+            {
+                PermissionController.hasFilePermission = false;
+            }
         }
 
         private bool DoesConfigFileExist()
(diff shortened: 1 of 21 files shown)


TF_Service_and_Tooling_Unity/Assets/TouchFree/ServiceShared/Scripts/Configuration/ConfigFiles/ConfigFile.cs
```diff
@@ -81,7 +81,14 @@ protected TData LoadConfig_Internal()
...
         protected void SaveConfig_Internal(TData config)
         {
-            File.WriteAllText(_ConfigFilePath, JsonUtility.ToJson(config, true));
+            try
+            {
+                File.WriteAllText(_ConfigFilePath, JsonUtility.ToJson(config, true));
+            }
+            catch
+            {
+                PermissionController.hasFilePermission = false;
```
NatGreenwood-UL, line 90: Because this only gets set when either creating a default config file, or saving a config file, does that mean that the popup won't appear until users try (and the app fails) to make a change or does something call save immediately and reveal the blockers? (I didn't have an easy way to test this on hand w/  my permissions chaos)

MattGrayUL: I added the new PermissionController.cs which does exactly that, it asks a config to try and save (which will load first because we access a config)

TF_Service_and_Tooling_Unity/Assets/TouchFree/ServiceUI/Scripts/Configuration/ConfigUI/InteractionSettingsConfigUI.cs
```diff
@@ -69,8 +69,18 @@ public class InteractionSettingsConfigUI : ConfigUI
         [Space, Tooltip("List all Settings elements that relate to the interactionType.")]
         public InteractionTypeElements[] interactionTypeElements;
 
+        public GameObject[] permissionsBlockers;
+
         private void Awake()
         {
+            if (!PermissionController.hasFilePermission)
```
PeterNancollis-UH: This process is repeated on multiple screens, could we put it in a helper or base class so we reuse one method?

TF_Service_and_Tooling_Unity/Assets/TouchFree/ServiceUI/Scripts/PermissionController.cs
```diff
@@ -0,0 +1,16 @@
...
+
+public class PermissionController : MonoBehaviour
+{
+    public static bool hasFilePermission = true;
+
+    // Force a save which causes the check for file access and if we do not
+    // have it, sets hasFilePermission to false
+    private void Awake()
+    {
+        ConfigManager.PhysicalConfig.SaveConfig();
```
PeterNancollis-UH, line 14: It feels a bit strange that we set `hasFilePermission` via a side effect of the `SaveConfig()`  method. Could `SaveConfig` return a bool or have an `out` parameter that sets the value so how this value gets set is clearly defined from this class?

MattGrayUL: I thought the same, but we also have other saves that do not need to know if it succeeded or not. I didn't think it was worth adding a specific function here to handle it and it happening upon every SaveConfig() feels like it's usefully double bagging it (also means we don't throw errors when it fails)

Returning a bool may be the nicest way to do this though to make it more explicit? I feel a out is overkill particularly when we don't always need to know the result

PeterNancollis-UH: I hadn't considered the benefit of this updating on every save. Given that I think this is fine as it is

PeterNancollis-UH (approved):

PR #155, opened 2021-09-15
TF-196 CTI files are accessible from Settings UI

MattGrayUL:
Moved CTI files to Settings UI app

- [x] Code Review


TF_Service_and_Tooling_Unity/Assets/TouchFree/ServiceShared/Scripts/Configuration/Managers/TFAppConfig.cs
```diff
@@ -52,7 +52,7 @@ public class ConfigurationState
 
         // CTI Settings
         public bool ctiEnabled = false;
-        public string ctiFilePath = Path.Combine(Application.streamingAssetsPath, "CallToInteract", "AirPush_Portrait.mp4");
+        public string ctiFilePath = Path.Combine(Application.streamingAssetsPath, "CallToInteract/AirPush_Portrait.mp4");
```
PeterNancollis-UH, line 55: Why is this change necessary? Isn't the result the same either way?

MattGrayUL: Well, both work as paths but it removes one of the double slashes we get in the config file, for some reason Path.Combine likes to add extra slashes

PeterNancollis-UH (approved):

PR #156, opened 2021-09-15
tf 170 touch free doesn't interact with settings UI

MattGrayUL:
Added a 1 second timer to check for TouchFreeService.exe being open. If it is, disables the client gameobject (which contains everything) disabling all functionality

- [x] Code Review

### DIFF
--- a/TF_Application/Assets/TouchFree_Application/Scripts/TouchFreeMain.cs
+++ b/TF_Application/Assets/TouchFree_Application/Scripts/TouchFreeMain.cs
@@ -14,16 +14,18 @@ namespace Ultraleap.TouchFree
         public static int CursorWindowSize = 200;
         public static Vector3 CursorWindowMiddle = new Vector2(100, 100);
 
-        bool settingsOpen = false;
-        public GameObject disableWhenSettingsOpen;
-
         static int minCursorWindowSize = 50;
         static int maxCursorWindowSize = 500;
+
+        bool settingsAppOpen = false;
+
+        public GameObject toolingClientGameobject;
+        public GameObject[] visibleCanvasses;
 
         void Awake()
         {
             Application.targetFrameRate = 60;
-            StartCoroutine(CheckForSettingsOpen());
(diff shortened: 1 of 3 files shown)


TF_Application/Assets/TouchFree_Application/Scripts/TouchFreeMain.cs
```diff
@@ -48,5 +58,39 @@ public static void UpdateCursorWindowSize(float _cursorSize)
...
+
+        IEnumerator CheckForSettingsOpen()
+        {
+            WaitForSeconds wait = new WaitForSeconds(1);
+
+            yield return wait;
+
+            while (true)
+            {
+                IntPtr hwnd = FindWindow(null, "TouchFreeService");
```
PeterNancollis-UH, line 72: This name will need to change after TF-193

MattGrayUL: I added a note to the miro board for this :)

TF_Application/Assets/TouchFree_Application/Scripts/TouchFreeMain.cs
```diff
@@ -48,5 +58,39 @@ public static void UpdateCursorWindowSize(float _cursorSize)
...
+
+                if ((int)hwnd != 0)
+                {
+                    if (!settingsOpen)
+                    {
+                        disableWhenSettingsOpen.SetActive(false);
```
PeterNancollis-UH: This would be better handled with a method that makes it clear what objects we are disabling/enabling.  At the moment I would need to open the scene to find that out, but ideally we should be able to follow the flow of the code without opening the scene.

PeterNancollis-UH:
Only issue I can find is if you push forward (triggering a down) then open the Settings UI. When you close the settings UI the TF cursor is still in a down state so you perform a drag until you trigger a new down event to clear it. This isn't likely to occur in normal use though as it's very hard to perform.

PeterNancollis-UH (approved):

PR #161, opened 2021-09-24
Stop setting desktop mode when exiting the Settings UI

PeterNancollis-UH:
- [x] Code Review


MattGrayUL (approved):
Tested locally against the known reproducible steps and it fixes the issue

PeterNancollis-UH:
This PR is no longer required as we will address the issue elsewhere

PR #162, opened 2021-09-24
Bug/tf 208 tracking mode reverts to desktop

MattGrayUL:
- [x] Code Review 

### DIFF
--- a/TF_Service_and_Tooling_Unity/Assets/Plugins/LeapMotion/Core/Scripts/LeapServiceProvider.cs
+++ b/TF_Service_and_Tooling_Unity/Assets/Plugins/LeapMotion/Core/Scripts/LeapServiceProvider.cs
@@ -53,7 +53,7 @@ namespace Leap.Unity {
 
     #region Inspector
     
-    public enum Mode
+    public enum TrackingMode
     {
       None,
       Desktop,
@@ -64,7 +64,7 @@ namespace Leap.Unity {
     [Tooltip("Sets the mode with which to initialise the tracking system. " +
       "The default value for this parameter is None, which implies that the tracking system " +
       "will remain in it's last known state. Change this to force a specific tracking mode on startup.")]
-    [SerializeField] private Mode _mode = Mode.None;
+    [SerializeField] private TrackingMode _trackingMode = TrackingMode.None;
 
     public enum InteractionVolumeVisualization {
       None,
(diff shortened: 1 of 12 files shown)


TF_Service_and_Tooling_Unity/Assets/Plugins/LeapMotion/Core/Scripts/LeapServiceProvider.cs
```diff
@@ -471,30 +473,110 @@ public enum TrackingOptimizationMode {
...
+        case Mode.None:
+          break;        
+        case Mode.Desktop:
+          _changeModeCoroutine = StartCoroutine(setMode(setDesktopModeFlags, mode));
+          break;
+        case Mode.ScreenTop:
+          _changeModeCoroutine = StartCoroutine(setMode(setScreenTopModeFlags, mode));
+          break;
+        case Mode.HeadMounted:
+          _changeModeCoroutine = StartCoroutine(setMode(setHeadMountedModeFlags, mode));
```
NatGreenwood-UL: Does setting this value when a coroutine is running kill the old coroutine or does it end up running headless? Concerned about having two competing with one another but not really sure how Unity Coroutines work regarding this.

MattGrayUL: This class is handled by XR team but to explain, the coroutine will never be running when a new one is started because it is always stopped at the beginning of this function, before this highlighted code executes

NatGreenwood-UL: Ah yeah I missed the Stop at the start. More for my Unity knowledge than for the sake of the code. I guess then that overwriting a coroutine _doesn't_ automatically stop the original? Good to know for future.

MattGrayUL: Indeed. You don't _have_ to store it as a coroutine, you can start them headlessly and they will just execute, but to handle the stopping of them you must store a reference. (or call StopAllCoroutines()).

As you guessed in your first comment, if we weren't calling StopCoroutine() above, they would all continue executing

PeterNancollis-UH (approved):

NatGreenwood-UL:
I've also now tested this locally with a build from this branch and the visualiser, I think we're good to merge it into the release branch now @PeterNancollis-UH ? I'll hold off on the release testing until XR have the module update approved on their end but I can't reproduce the bug any more so I think that justifies this being in the release so we're a step closer to ready?

PeterNancollis-UH:
@NatGreenwood-UL Agreed, I will merge it now

PR #168, opened 2021-10-12
Feature/tf 220+tf 215 service configuration dot net

MattGrayUL:
- [x] Code Review 


TF_Service_dotNet/TouchFree Service/Configuration/ConfigManager.cs
```diff
@@ -0,0 +1,68 @@
+﻿using System;
+
+namespace Ultraleap.TouchFree.Service.Configuration
+{
+    public static class ConfigManager
+    {
+        //public static GlobalSettings GlobalSettings
```
PeterNancollis-UH, line 7: Why is this section commented out? if it's not being used can it be removed?

TF_Service_dotNet/TouchFree Service/Program.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+
+namespace Ultraleap.TouchFree.Service
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            Console.WriteLine(Configuration.ConfigManager.PhysicalConfig.ScreenHeightM);
```
PeterNancollis-UH: Probably not worth worrying about as this won't persist in the service, but would be useful to identify what this value is in the message. A seemingly number dumped to the console is a bit cryptic

.gitlab-ci.yml
```diff
@@ -16,376 +16,25 @@ variables:
...
-    - cd ./Scripts/
-    - "& ./compile_installer.bat ${env:TOUCHFREE_VERSION} $CI_PROJECT_DIR/TF_Service_Utilities/Installer/TouchFree_Installer.iss"
-
-    # Sign the installer with the UltraLeap certificate
-    - $cert=Get-ChildItem -Path Cert:\LocalMachine\My -CodeSigningCert
-    - Set-AuthenticodeSignature -FilePath "$CI_PROJECT_DIR/Installer_Build/TouchFree_${TOUCHFREE_VERSION}_Installer.exe" -TimestampServer 'http://timestamp.comodoca.com/authenticode' -Certificate $cert
+    - Set-AuthenticodeSignature -FilePath "$CI_PROJECT_DIR/Service_dotNet_Build/TouchFree Service.exe" -TimestampServer 'http://timestamp.comodoca.com/authenticode' -Certificate $cert
   artifacts:
-    name: "TouchFree_Installer_${TOUCHFREE_VERSION}_${CI_COMMIT_SHORT_SHA}"
+    name: "TouchFree_Service_dotNet"
```
PeterNancollis-UH, line 149: The build artefact contains two copies of the resulting exe, on in the root folder and one in the publish folder. I presume we don't really need both

.gitlab-ci.yml
```diff
@@ -16,376 +16,25 @@ variables:
...
-  - test
-  - package
-  - installer
 
-TF_Tooling::Unity::build:
+TF_Service::dotNet::build:
   stage: build
   when: always
   script:
-    # Get Unity version and changeset from the project.
```
PeterNancollis-UH: Removing all the build steps we use on develop feels like it makes work to put it all back in future when we come to merge the epic branch. Would it make life easier later if we leave the code in the file and just disable the stages we don't want during the development process?

.gitlab-ci.yml
MattGrayUL: We do the same on develop for all artifacts, but we can change the 'copy' to a 'move' if we want I assume?

TF_Service_dotNet/TouchFree Service/Configuration/ConfigManager.cs
MattGrayUL: It exists for the virtual screen system in the Unity service so may be useful to keep for when that work begins. However I think once that work is completed we should remove the vague 'GlobalSettings' object

PeterNancollis-UH: If we will need it in the near future then it's fine to leave it

.gitlab-ci.yml
PeterNancollis-UH: Looking at it again it might be that we only need to copy the contents of the publish directory, whereas currently it copies the parent

PeterNancollis-UH (approved):

PR #169, opened 2021-10-13
TF-225 added Leap csharp into project

MattGrayUL:
Added leapc# to the project, copies leapc.dll to build directory and logs the palm position to console

- [x] Code Review 


PeterNancollis-UH:
The copy of LeapCSharp still contains all the Unity meta files. I don't think we need to keep these in the project so should we remove them?

.gitlab-ci.yml
```diff
@@ -138,9 +138,9 @@ TF_Service::dotNet::build:
   when: always
   script:
     # Move the built wrapper to the root directory
-    - dotnet publish -c Release "$CI_PROJECT_DIR/TF_Service_dotNet/TouchFree Service/TouchFree Service.sln"
+    - dotnet build -c Release "$CI_PROJECT_DIR/TF_Service_dotNet/TouchFree Service/TouchFree Service.sln"
```
PeterNancollis-UH, line 141: Why have we changed back to a build instead of publish?

TF_Service_dotNet/TouchFree Service/Program.cs
```diff
@@ -1,12 +1,34 @@
...
 
 namespace Ultraleap.TouchFree.Service
 {
     class Program
     {
+        private Controller controller = new Controller();
+
         static void Main(string[] args)
+        {
+            new Program().Run();
```
PeterNancollis-UH, line 12: I've not seen this pattern before, seems strange to create a new instance of a class from within main that's defined in that class. Is this a normal way of working do you know?

.gitlab-ci.yml
MattGrayUL: The copying of the LeapC.dll needs to happen on a build (it is ignored on a publish). I also realised the intended route to publishing is to build first and then publish after if you need to. However we actually don't _need_ to publish.

Publishing was something I left in as it didn't cause issues at the time and would allow for an easier 'also include runtimes' solution if we found we needed it.

TF_Service_dotNet/TouchFree Service/Program.cs
MattGrayUL: I don't know if it is a normal way to work, however while we do not have the basic loop set up yet, I cannot create non-static functions that are called from Main, so I copied this pattern from another project

PeterNancollis-UH: The `Run` function here is establishing that basic loop, so we could pop that functionality into a class of it's own rather than it being in `Program`. Main could then instance that class and call run. 

This could be something that could be resolved in the next task as we start to introduce the functional parts of the program

PeterNancollis-UH (approved):

PR #170, opened 2021-10-13
TF-242 set tracking mode on launch

MattGrayUL:
Created new TrackingModeManager for setting modes. Used TrackingModeManager on launch

- [ ] Code Review 

### DIFF
--- a/TF_Service_dotNet/TouchFree Service/Program.cs
+++ b/TF_Service_dotNet/TouchFree Service/Program.cs
@@ -5,7 +5,7 @@ namespace Ultraleap.TouchFree.Service
 {
     class Program
     {
-        private Controller controller = new Controller();
+        public static Controller controller = new Controller();
 
         static void Main(string[] args)
         {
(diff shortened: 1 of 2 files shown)


TF_Service_dotNet/TouchFree Service/TrackingModeManager.cs
```diff
@@ -0,0 +1,57 @@
...
+            }
+            else
+            {
+                SetTrackingMode(TrackingMode.DESKTOP);
+            }
+        }
+
+        static void SetTrackingMode(TrackingMode _mode)
+        {
+            Console.WriteLine($"Requesting {_mode} tracking mode");
```
PeterNancollis-UH, line 31: We should consider introducing a mechanism to redirect these messages to a log file as once we move to a service this will not be visible.

TF_Service_dotNet/TouchFree Service/Program.cs
```diff
@@ -5,7 +5,7 @@ namespace Ultraleap.TouchFree.Service
 {
     class Program
     {
-        private Controller controller = new Controller();
+        public static Controller controller = new Controller();
```
PeterNancollis-UH, line 8: Seeing how the Program class is expanding and static elements are being introducing I think we might want to take a moment to plan the basic architecture of the app so we don't end up having to come back and rework it when there's more going on

TF_Service_dotNet/TouchFree Service/TrackingModeManager.cs
MattGrayUL: The wrapper has a log file so as long as what we log is picked up by the wrapper, it should go to the file?

PR #178, opened 2021-10-18
TF-136 Run unity tests on ci

PeterNancollis-UH:
- [x] Code Review
- [x] Developer Testing

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -161,7 +161,7 @@ TF_Service::dotNet::test:
   when: always
   script:
     # Run unit tests
-    - dotnet test "$CI_PROJECT_DIR/TF_Service_dotNet/TouchFreeTests/TouchFreeTests.csproj" --logger:"junit;LogFilePath=$CI_PROJECT_DIR/TestResults/Test_Results.xml;"
+    - dotnet test "$CI_PROJECT_DIR/TF_Service_dotNet/TouchFreeTests/TouchFreeTests.csproj" --logger:"junit;LogFilePath=$CI_PROJECT_DIR/TestResults/Test_Results.xml;MethodFormat=Class;FailureBodyFormat=Verbose"
 
   artifacts:
     when: always
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -144,7 +144,7 @@ TF_Service::dotNet::build:
 
     # Move the built executable to the root directory
     - mkdir ./Service_dotNet_Build
-    - cp -r "$CI_PROJECT_DIR/TF_Service_dotNet/TouchFree Service/bin/Release/net5.0/publish/*" ./Service_dotNet_Build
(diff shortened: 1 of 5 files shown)


.gitlab-ci.yml
```diff
@@ -137,10 +137,14 @@ TF_Service::dotNet::build:
   stage: build
   when: always
   script:
-    # Move the built wrapper to the root directory
-    - dotnet publish -c Release "$CI_PROJECT_DIR/TF_Service_dotNet/TouchFree Service/TouchFree Service.sln"
+    - dotnet build -c Release "$CI_PROJECT_DIR/TF_Service_dotNet/TouchFree Service/TouchFree Service.sln"
+
+    # Run unit tests
+    - dotnet test "$CI_PROJECT_DIR/TF_Service_dotNet/TouchFreeTests/TouchFreeTests.csproj"
```
NatGreenwood-UL: We've got a `test` stage in the main CI atm, would it be cleaner to just build the executables here and test them in the test stage? Also, I'm wondering if we could get the unit tests to produce a results xml file, because we could capture that with [GitLab's results](https://docs.gitlab.com/ee/ci/unit_test_reports.html), which may (hopefully) pass the same info to GitHub. It doesn't really matter now, but if we can trivially set up clean test reports now, it may save us a bunch of digging through GitLab consoles to find out what failed later 😅

.gitlab-ci.yml
```diff
@@ -137,10 +137,14 @@ TF_Service::dotNet::build:
...
+
+    # Move the built executable to the root directory
     - mkdir ./Service_dotNet_Build
-    - cp -r "$CI_PROJECT_DIR/TF_Service_dotNet/TouchFree Service/bin/Release/net5.0/publish/*" ./Service_dotNet_Build
+    - cp -r "$CI_PROJECT_DIR/TF_Service_dotNet/TouchFree Service/bin/Release/net5.0/*" ./Service_dotNet_Build
```
NatGreenwood-UL, line 144: I believe @MattGrayUL was publishing the build so it included a .NET5 runtime, did we figure out another way around that?

PeterNancollis-UH: I switched to doing just a build as @MattGrayUL found that the publish didn't work as intended. We would need to do more work to include the runtime. The publish also caused some trouble when trying to add the LeapC.dll to the project, so this is partly in preparation for other work

.gitlab-ci.yml
PeterNancollis-UH: Happy to look at splitting it into two stages and to see if I can get the xml. Might as well get it right first time

PeterNancollis-UH: Have made these changes. The results are now output as xml, which displays in the test summary in GitLab

NatGreenwood-UL: Looks great, as a note, you can also now navigate reasonably easily to those results from right here on a GitHub PR by expanding the checks below and following the link to GitLab! Also a big fan of the results summary screen on GitLab: 
image

NatGreenwood-UL (approved):
Looking great with the reports working! Now to just write those Unit tests for the Library 😅

PR #179, opened 2021-10-19
TF-213/TF-216 WebSocket API in raw ASP.NET

NatGreenwood-UL:
- [x] Code Review 

### DIFF
--- a/TF_Service_dotNet/TouchFree Service/Connection/ClientConnection.cs
+++ b/TF_Service_dotNet/TouchFree Service/Connection/ClientConnection.cs
@@ -57,6 +57,7 @@ namespace Ultraleap.TouchFree.Service.Connection
 
             string jsonMessage = JsonConvert.SerializeObject(message);
 
+            Send(jsonMessage);
         }
 
         public void SendHandshakeResponse(ResponseToClient _response)
--- a/TF_Service_dotNet/TouchFree Service/Connection/ClientConnection.cs
+++ b/TF_Service_dotNet/TouchFree Service/Connection/ClientConnection.cs
@@ -17,14 +17,14 @@ namespace Ultraleap.TouchFree.Service.Connection
         public WebSocket socket;
         private bool HandshakeCompleted;
         private readonly WebSocketReceiver receiver;
+        private readonly ClientConnectionManager clientMgr;
 
-        public ClientConnection(WebSocket _socket, WebSocketReceiver _receiver)
+        public ClientConnection(WebSocket _socket, WebSocketReceiver _receiver, ClientConnectionManager _clientMgr)
         {
             socket = _socket;
             receiver = _receiver;
+            clientMgr = _clientMgr;
             HandshakeCompleted = false;
-
-            //socket.ReceiveAsync
 
             Console.WriteLine("Websocket Connection opened");
         }
@@ -102,7 +102,7 @@ namespace Ultraleap.TouchFree.Service.Connection
 
         private void SendInitialHandState()
         {
-            this.SendHandPresenceEvent(ClientConnectionManager.Instance.missedHandPresenceEvent);
+            this.SendHandPresenceEvent(clientMgr.missedHandPresenceEvent);
         }
 
         private Compatibility GetVersionCompability(string _clientVersion, Version _coreVersion)
--- a/TF_Service_dotNet/TouchFree Service/Connection/ClientConnection.cs
+++ b/TF_Service_dotNet/TouchFree Service/Connection/ClientConnection.cs
@@ -183,7 +183,7 @@ namespace Ultraleap.TouchFree.Service.Connection
                 // Validation has failed because there is no valid requestID
                 response.status = "Failure";
                 response.message = "Handshaking failed. This is due to a missing or invalid requestID";
-                Console.Error.WriteLine("Handshaking failed. This is due to a missing or invalid requestID");
+                Console.Error.WriteLine(response.message);
                 SendHandshakeResponse(response);
                 return;
             }
@@ -196,7 +196,7 @@ namespace Ultraleap.TouchFree.Service.Connection
                 // cannot be processed
                 response.status = "Failure";
                 response.message = "Request Rejected: Requests cannot be processed until handshaking is complete.";
-                Console.Error.WriteLine("Request Rejected: Requests cannot be processed until handshaking is complete.");
+                Console.Error.WriteLine(response.message);
                 SendHandshakeResponse(response);
                 return;
             }
@@ -206,7 +206,7 @@ namespace Ultraleap.TouchFree.Service.Connection
                 // Send back immediate error: Cannot compare version number w/o a version number
                 response.status = "Failure";
                 response.message = "Handshaking Failed: No API Version supplied.";
-                Console.Error.WriteLine("Handshaking Failed: No API Version supplied.");
+                Console.Error.WriteLine(response.message);
                 SendHandshakeResponse(response);
                 return;
             }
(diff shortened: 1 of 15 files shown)


TF_Service_dotNet/TouchFree Service/Connection/ClientConnection.cs
```diff
@@ -0,0 +1,241 @@
...
+
+        public void SendHandPresenceEvent(HandPresenceEvent _response)
+        {
+            CommunicationWrapper<HandPresenceEvent> message =
+                new CommunicationWrapper<HandPresenceEvent>(
+                    ActionCode.HAND_PRESENCE_EVENT.ToString(),
+                    _response);
+
+            string jsonMessage = JsonConvert.SerializeObject(message);
+
```
PeterNancollis-UH, line 59: Just so it doesn't get missed, this one seems to be missing it's `Send(jsonMessage);`

NatGreenwood-UL: Dropped this in!

TF_Service_dotNet/TouchFree Service/Program.cs
```diff
@@ -1,29 +1,27 @@
-﻿using System;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
```
MattGrayUL: Are these needed?

TF_Service_dotNet/TouchFree Service/Program.cs
```diff
@@ -1,29 +1,27 @@
...
+
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.Hosting;
+
 using Ultraleap.TouchFree.Library.Configuration;
```
MattGrayUL: Is this needed?

TF_Service_dotNet/TouchFree Service/UpdateBehaviour.cs
```diff
@@ -7,7 +7,7 @@
```
MattGrayUL, line 2: Only the System.Timers Using is actually used, the others could be removed?

TF_Service_dotNet/TouchFree Service/Connection/Middlewares/TouchFreeRouter.cs
```diff
@@ -0,0 +1,76 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+using Microsoft.AspNetCore.Http;
+
+namespace Ultraleap.TouchFree.Service.Connection
+{
+    public class TouchFreeRouter
+    {
+        private readonly RequestDelegate next;
```
MattGrayUL, line 13: What is a 'next'?

TF_Service_dotNet/TouchFree Service/Connection/Middlewares/TouchFreeRouter.cs
```diff
@@ -0,0 +1,76 @@
...
+
+                        Console.WriteLine("Websocket Connection Closed");
+
+                        return;
+                    }
+                });
+            }
+            else
+            {
+                Console.WriteLine("A request was made to the server that was not a ");
```
MattGrayUL: Is this sentence incomplete?

TF_Service_dotNet/TouchFree Service/Connection/Middlewares/TouchFreeRouterExtensions.cs
```diff
@@ -0,0 +1,47 @@
...
+        {
+            services.AddSingleton<UpdateBehaviour>();
+            return services;
+        }
+
+        public static IServiceCollection AddConfigFileWatcher(this IServiceCollection services)
+        {
+            var watcher = new ConfigFileWatcher();
+
+            services.BuildServiceProvider().GetService<UpdateBehaviour>().OnUpdate += watcher.Update;
```
MattGrayUL, line 39: Should this be done via the constructor of ConfigFileWatcher (like others) rather than directly?

TF_Service_dotNet/TouchFree Service/Connection/ClientConnection.cs
```diff
@@ -178,7 +183,7 @@ protected void ProcessHandshake(ActionCode action, string requestContent)
                 // Validation has failed because there is no valid requestID
                 response.status = "Failure";
                 response.message = "Handshaking failed. This is due to a missing or invalid requestID";
-                Debug.LogError("Handshaking failed. This is due to a missing or invalid requestID");
+                Console.Error.WriteLine("Handshaking failed. This is due to a missing or invalid requestID");
```
MattGrayUL: I know it was like this previously, but can we not use the response.message reference as the console error? that way we only have on string to edit if we edit it?

TF_Service_dotNet/TouchFree Service/Connection/Services/WebSocketReceiver.cs
```diff
@@ -1,21 +1,30 @@
...
-    public class WebSocketReceiver : MonoBehaviour
+    public class WebSocketReceiver
     {
         public ConcurrentQueue<string> configChangeQueue = new ConcurrentQueue<string>();
         public ConcurrentQueue<string> configStateRequestQueue = new ConcurrentQueue<string>();
 
+        private readonly UpdateBehaviour updateBehaviour;
+        private readonly ClientConnectionManager clientMgr;
+
+        public WebSocketReceiver(UpdateBehaviour _updateBehaviour, ClientConnectionManager _clientMgr)
```
MattGrayUL, line 20: Is this ever constructed?

TF_Service_dotNet/TouchFree Service/Connection/Services/WebSocketReceiver.cs
```diff
@@ -225,7 +235,8 @@ void ChangeConfig(string _content)
         {
             ConfigState combinedData = new ConfigState("", ConfigManager.InteractionConfig, ConfigManager.PhysicalConfig);
 
-            JsonUtility.FromJsonOverwrite(_content, combinedData);
+            JsonConvert.PopulateObject(_content, combinedData);
+            //JsonUtility.FromJsonOverwrite(_content, combinedData);
```
MattGrayUL: no need for this line now?

MattGrayUL (commented):
A few comments here, mostly just little adjustments though. I have not run this locally but thought i'd get a code review in now 😄

TF_Service_dotNet/TouchFree Service/Connection/Middlewares/TouchFreeRouter.cs
NatGreenwood-UL: In `Configure` in `Startup` you can see we add the TouchFreeRouter as a Middleware, but you can add as many of these as you'd like and incoming requests will be passed through the chain. The "next" delegate is what we fire when this middleware has done what it needs and wants to pass it onwards through the chain.

Here's an image from [some tutorials I've seen linked from various places ](https://www.tutorialsteacher.com/core/aspnet-core-middleware) that might make it clearer
image

TF_Service_dotNet/TouchFree Service/Connection/Middlewares/TouchFreeRouter.cs
NatGreenwood-UL: It **should** read "A request was made to the server that was not an attempt to conenct to a WebSocket?" 😅 Fix coming

TF_Service_dotNet/TouchFree Service/Connection/Middlewares/TouchFreeRouterExtensions.cs
NatGreenwood-UL: I handled this one differently because the ConfigFileWatcher is the only one instanced here that lives in the Library instead of in the Service. If I made this change to the ConfigFileWatcher's constructor, there will be a dependency on the Service code within the Library. I figure that we want the library to be able to stand alone if we can?

MattGrayUL: Yeah that makes sense to me then :)

TF_Service_dotNet/TouchFree Service/Connection/Services/WebSocketReceiver.cs
NatGreenwood-UL: Yup, but the reference isn't shown in VS because of the roundabout instancing, see `AddWebSocketReceiver` in `TouchFreeRouterExtensions` (which itself is called in `Startup`'s `ConfigureServices` method)

MattGrayUL: I noticed that but how is that given an UpdateBehaviour or a ClientConnectionManager? Is that because they are singletons?

MattGrayUL (approved):
Tested locally and an existing client can connect with expected handshake response messages

TF_Service_dotNet/TouchFree Service/Connection/Services/WebSocketReceiver.cs
NatGreenwood-UL: Yeah so if something being constructed as a Middleware or service has something in its constructor, it will look to the existing Singleton services to pass in. This is why the initialisation of the WSReciever has to be the _last_ of the services added in the Services list, cos it depends on the other ones.

PR #180, opened 2021-10-20
Feature/tf 217/tf 242 tracking mode set on launch

MattGrayUL:
- [x] Code Review 


TF_Service_dotNet/TouchFree/TrackingConnectionManager.cs
```diff
@@ -0,0 +1,65 @@
+﻿using System;
+using Leap;
+using Ultraleap.TouchFree.Library.Configuration;
+
+namespace Ultraleap.TouchFree.Library
+{
+    public class TrackingConnectionManager
+    {
+        Controller controller;
```
NatGreenwood-UL: Given the WebSocket troubles we ran into yesterday, it may be worth specifying this as a (Leap?) controller here to avoid any confusion in integration (I know this is internal to the Service but it **is** opensource)

TF_Service_dotNet/TouchFree Service/Program.cs
```diff
@@ -11,6 +12,8 @@ static void Main(string[] args)
             UpdateBehaviour updateLoop = new();
             updateLoop.OnUpdate += TickTock;
 
+            TrackingConnectionManager trackingConnectionManager = new TrackingConnectionManager();
```
NatGreenwood-UL: I think this will collide with my other PR as we'll need to instance this as an ASP Service. Not an issue but we should note that whichever gets merged second will need to handle the changes from the first.

NatGreenwood-UL (approved):
Have tested this locally and it works with the expected port and Leap, including setting the tracking mode in the log successfully!

PR #181, opened 2021-10-20
Feature/tf 217/tf 241 virtual screen for dot net

MattGrayUL:
- [x] Code Review 


NatGreenwood-UL:
The code itself looks good, but where is it going to be used / what is going to own the VirtualScreen instance?

MattGrayUL:
> The code itself looks good, but where is it going to be used / what is going to own the VirtualScreen instance?

The VirtualScreen can be instanced via its static 'virtualScreen' reference which will generate a new one if one does not exist. It will be referenced by the interaction code when to convert the desired point on the hand into screen space.

I can add an example piece of code to log one based on an arbitrary point on the hand if it helps with testing/proving, part of the main TF-217 task is to also send arbitrary data over the websocket so it will be required to complete the overarching task anyway

The VirtualScreen will be better used when the handmanager and interaction code is implemented.

PeterNancollis-UH (approved):

PR #183, opened 2021-10-27
Feature/tf 218 hand manager in dot net

PeterNancollis-UH:
- [x] Code Review 

### DIFF
--- a/TF_Service_dotNet/TouchFree/HandManager.cs
+++ b/TF_Service_dotNet/TouchFree/HandManager.cs
@@ -234,11 +234,6 @@ namespace Ultraleap.TouchFree.Library
             }
         }
 
-        public bool IsLeapServiceConnected()
-        {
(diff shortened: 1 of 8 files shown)


TF_Service_dotNet/TouchFree/HandManager.cs
```diff
@@ -286,42 +209,39 @@ void AssignNewPrimary(Hand _left, Hand _right)
...
-                secondaryChirality = Chirality.Left;
+                secondaryChirality = HandChirality.LEFT;
             }
         }
 
         public bool IsLeapServiceConnected()
         {
-            return trackingProvider.IsConnected();
+            return trackingProvider.controller.IsConnected;
         }
```
MattGrayUL: I believe this was only used for the settings UI so could be removed?

MattGrayUL (approved):

PR #184, opened 2021-10-29
Feature/tf 219 position stabiliser in dot net

MattGrayUL:
- [x] Code Review 


TF_Service_dotNet/TouchFree/PositionStabiliser.cs
```diff
@@ -0,0 +1,149 @@
...
+            // Assumes _progressToClick is clamped.
+            var scaledValue = _progressToClick * _progressToClick;
+            var deadZoneRadius = defaultDeadzoneRadius * scaledValue;
+
+            currentDeadzoneRadius = deadZoneRadius;
+        }
+
+        void OnSettingsUpdated(BaseConfig _baseConfig)
+        {
+            defaultDeadzoneRadius = ((InteractionConfig)_baseConfig).DeadzoneRadius;
```
PeterNancollis-UH: This is a different style of casting the BaseConfig to the specific type than was used for the hands module. Do we have a preferred option we can agree to use?

TF_Service_dotNet/TouchFree Service/Connection/Middlewares/TouchFreeRouterExtensions.cs
```diff
@@ -56,5 +56,11 @@ public static IServiceCollection AddHandManager(this IServiceCollection services
             services.AddSingleton<HandManager>();
             return services;
         }
+
+        public static IServiceCollection AddPositionStabiliser(this IServiceCollection services)
+        {
+            services.AddSingleton<PositionStabiliser>();
```
PeterNancollis-UH, line 64: I wonder if this needs to be a singleton once we add the Interaction Manager given it's only the interactions that make use of it. Doesn't need changing now necessarily but something to consider

TF_Service_dotNet/TouchFree/PositionStabiliser.cs
MattGrayUL: Happy to adjust to the way it is in HandManager, it is more readable

`PhysicalConfig config = _config as PhysicalConfig;`

TF_Service_dotNet/TouchFree Service/Connection/Middlewares/TouchFreeRouterExtensions.cs
MattGrayUL: The way it exists in Unity is as a single object that is accessed by all of the PositioningModules (which are attached to the individual Interactions)

We _could_ use one instance per interaction, but it is more of a utility class for the PositioningModules. Perhaps we pass it as a reference to the interactions which can then pass it to their respective PositioningModule?

PeterNancollis-UH: Yes, something like that. I'm thinking the Interaction manager will be responsible for instantiating the interactions and their dependencies, so it could manage this without the need for the singleton position stabiliser

MattGrayUL: I agree, shall I comment it to mark it for deletion in future but leave it in for testing for now?

PeterNancollis-UH (approved):

PR #185, opened 2021-10-29
Bug adjust non-static events for ConfigUpdated on BaseConfig

MattGrayUL:
Converted non-static events for ConfigUpdated on BaseConfig to exist in the specific Interactions as static events


PeterNancollis-UH:
A couple of alternative solutions to this problem come to mind that don't require static references to events:

- We could stop replacing the PhysicalConfig object in the ConfigManager when we get new data and instead just update the values inside the same object. That way the event links will be preserved.
- We could also shift responsibility to informing on config changes to the config manager, leaving the configs themselves to be just data objects. In this case we could replace the config as we currently do and the event links are preserved via the manager. I think this might be an option I prefer as it provides better separation of responsibility

MattGrayUL:
> * We could also shift responsibility to informing on config changes to the config manager, leaving the configs themselves to be just data objects. In this case we could replace the config as we currently do and the event links are preserved via the manager. I think this might be an option I prefer as it provides better separation of responsibility

I like this one. I think we should do this

PeterNancollis-UH:
I'm also finding that if I try to update the physical config I don't see the change reflected in the cursor position. How did you test this?

MattGrayUL:
> I'm also finding that if I try to update the physical config I don't see the change reflected in the cursor position. How did you test this?

It _should_... the HandManager uses the static event that should be triggered by a file change

PeterNancollis-UH:
> > I'm also finding that if I try to update the physical config I don't see the change reflected in the cursor position. How did you test this?
> 
> It _should_... the HandManager uses the static event that should be triggered by a file change

It does in fact work in the Hand Manager, I was changing values the hand manager doesn't use. The virtual screen doesn't listen to the config changed events though so we should add that so it also receives updated config

TF_Service_dotNet/TouchFree/VirtualScreen.cs
```diff
@@ -132,14 +147,6 @@ public Vector2 MetersToPixels(Vector2 position)
             return positionInPixels;
         }
 
-        public static void CaptureCurrentResolution()
```
PeterNancollis-UH, line 135: Should this have been removed? Can see why it shouldn't be doing what it does in the virtual screen, but I think we still need to capture the resolution of the screen.

MattGrayUL: We shouldn't ever need to capture it now as it should always be set by the Settings?

The only benefit we get with having it auto-capture (which this currently isn't) is if the user changes resolution _since_ the last configuration, but capturing resolutions on application launch is in my experience unreliable so requires an arbitrary delay?

PeterNancollis-UH: That makes sense, just wanted to check

PeterNancollis-UH (approved):

PR #186, opened 2021-11-02
Feature/tf 187 masking sliders

MattGrayUL:
- [x] Code Review 


TF_Service_and_Tooling_Unity/Assets/TouchFree/ServiceUI/Scripts/DiagnosticAPI.cs
```diff
@@ -0,0 +1,227 @@
...
+    public uint connectedDeviceID;
+    public bool maskingAllowed = false;
+
+    const string minimumMaskingAPIVerison = "2.0.0";
+
+    ConcurrentQueue<string> newMessages = new ConcurrentQueue<string>();
+
+    public DiagnosticAPI(MonoBehaviour _creatorMonobehaviour)
+    {
+        Debug.Log("DiagnosticAPI constructor... ");
```
NatGreenwood-UL: Do we need this alert in generally / for testing purposes? If the latter (or not needed) we should probably knock it out before merge. Same goes for a string of Debugs below

TF_Service_and_Tooling_Unity/Assets/TouchFree/ServiceUI/Scripts/DiagnosticAPI.cs
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+    void IDisposable.Dispose ()
+    {
+        status = Status.Expired;
+        webSocket.Close();
+    }
+
+    [Serializable]
+    public struct ImageMaskData
```
NatGreenwood-UL: I'm used to keeping files to a single-class so my first impulse would be to shift these structs out to either an existing type file or a new one, but I noticed that these structs are all declared _inside the class_, does that have any benefits? Does it actually scope them to being only available to the class? If so I'm wondering if we could drop them up at the top above the constructor/member declarations? My general preference is to have the accessories at the top and then get into the meat of the class at the bottom, though if they're _inside_ the type that doesn't seem to work as well. 

* If there's benefit to them being declared _within_ the class I think this is fine
* if there _isn't_ I'd prefer to see them shifted outside of the class and at the top
* if they're referenced outside of this class, that's when I'd prefer them to be out in a types-style file, rather than having to dig to the bottom of this class to find a type to understand it when coming from another class entirely.

TF_Service_and_Tooling_Unity/Assets/TouchFree/ServiceUI/Scripts/DiagnosticAPI.cs
MattGrayUL: Yeah very true, there's probably a few of these logs we can remove

TF_Service_and_Tooling_Unity/Assets/TouchFree/ServiceUI/Scripts/DiagnosticAPI.cs
MattGrayUL: These structs are specifically utility for the DiagnosticAPI. I believe so long as we keep their use strictly within the DiagnosticAPI class, it makes sense?

Perhaps I should convert them from being public to ensure they are not used externally.

In doing so I should convert the OnGetMaskingResponse event that is triggered and used via CameraPreviewScreen.cs away from using one of the internal structs

NatGreenwood-UL:
I've tested this by modifying the version number locally and trying it out in-Editor, but I can't prove it out with even a development version of the tracking as the Diagnostic API version hasn't been iterated. Once we have a pre-release version with the API change we can run the tests in the test plan for merge.

NatGreenwood-UL (approved):
All tested locally with 5.2.0 and 5.3.0 and works well, looks great now with the change to check the API version on launch as well 👍

PR #187, opened 2021-11-03
TF-136/TF-223 Touch Plane interaction in .NET

NatGreenwood-UL:
- [x] Code Review 
- [ ] Developer Testing: Execution of the Touch Plane test plan


TF_Service_dotNet/TouchFree Service/Startup.cs
```diff
@@ -27,65 +22,25 @@ public void ConfigureServices(IServiceCollection services)
...
             app.UseWebSockets();
-
             app.UseTouchFreeRouter();
 
-            clientConnectionManager = app.ApplicationServices.GetService<ClientConnectionManager>();
-            positionStabiliser = app.ApplicationServices.GetService<PositionStabiliser>();
-            handManager = app.ApplicationServices.GetService<HandManager>();
-            handManager.HandsUpdated += UpdateHand;
-            app.ApplicationServices.GetService<UpdateBehaviour>().OnUpdate += Update;
+            interactionManager = app.ApplicationServices.GetService<InteractionManager>();
```
MattGrayUL: Do we actually need this anymore? will anything need this reference after this point or can it just exist as a service without a reference?

TF_Service_dotNet/TouchFree/Interactions/InteractionModules/InteractionModule.cs
```diff
@@ -1,41 +1,51 @@
...
         private HandChirality handChirality;
         public HandType handType;
 
         public bool ignoreDragging;
-        public PositioningModule positioningModule;
 
-        public delegate void InteractionInputAction(HandChirality _chirality, HandType _handType, InputAction _inputData);
-        public static event InteractionInputAction HandleInputAction;
+        public delegate void InteractionInputAction(InputAction _inputData);
+        public event InteractionInputAction HandleInputAction;
```
MattGrayUL: This could actually have the delegate removed and become:

`public event Action<InputAction> HandleInputAction;`

TF_Service_dotNet/TouchFree/Interactions/PositioningModule.cs
```diff
@@ -1,32 +1,32 @@
...
+            Stabiliser = _stabiliser;
+            trackedPosition = _position;
 
+            Enable();
+        }
+
+        protected void Enable()
+        {
             Stabiliser.ResetValues();
         }
```
MattGrayUL: This could be in one line rather than having an Enable method?

also do we even need to reset the values if each interaction has its own PositionStabiliser?

TF_Service_dotNet/TouchFree/ServiceTypes.cs
```diff
@@ -180,6 +182,22 @@ public static class Utilities
 
             return returnVal;
         }
+
+        public static System.Numerics.Vector3 LeapToNumerics(Leap.Vector _leap)
```
MattGrayUL: Should this specify in the method name that is produces a Vector3?

MattGrayUL (commented):
We should move the PositionStabiliser.cs into the Interaction folder and namespace

TF_Service_dotNet/TouchFree Service/Startup.cs
NatGreenwood-UL: Without this reference it won't get instanced by anything (as services are only instanced when something asks after them). I could try not _storing_ the reference and just getting the InteractionManager as an empty statement?

TF_Service_dotNet/TouchFree/ServiceTypes.cs
NatGreenwood-UL: My thought was that (if need be) we could overload this method to handle Vector2s as well so you can just call `LeapToNumerics` for the relevant type without having to specify the type explicitly in the function name.

TF_Service_dotNet/TouchFree Service/Startup.cs
MattGrayUL: Ah yeah I remember that now! If storing the reference isn't required, let's just remove that then :)

TF_Service_dotNet/TouchFree/ServiceTypes.cs
MattGrayUL: I think that makes sense though I think for clarity it should be LeapToNumericsVector in that case then? I guess it's personal preference but I would prefer the name not seem so broad, 'Leap' covers a lot of types and so does 'Numerics'

TF_Service_dotNet/TouchFree/Interactions/PositioningModule.cs
NatGreenwood-UL: It's pulled out in the enable method because it will need to be called on each instance of switching between interactions, not _just_ on initialisation (which I admittedly haven't done any set-up of within TouchPlane (Enable/Disable behaviour) yet).

Also I _think_ the Stabiliser needs to be reset even with seperate Stabilizers because we don't want to try to stabilise new data based on old? Not sure it'll have a _huge_ effect as the stabilizer only caches one previous pos but it could cause some funky drift from the last position this interaction saw a hand to the actual hand pos if we don't reset.

MattGrayUL (approved):

PR #188, opened 2021-11-05
TF-136/TF-221 air push interaction in dot net

NatGreenwood-UL:
- [x] Code Review 

### DIFF
--- a/TF_Service_dotNet/TouchFree/Interactions/InteractionModules/AirPushInteraction.cs
+++ b/TF_Service_dotNet/TouchFree/Interactions/InteractionModules/AirPushInteraction.cs
@@ -1,0 +1,326 @@
+﻿using UnityEngine;
+using Stopwatch = System.Diagnostics.Stopwatch;
+using System.Collections.Generic;
+using Ultraleap.TouchFree.ServiceShared;
+
+namespace Ultraleap.TouchFree.Service
+{
+    public class AirPushInteraction : InteractionModule
+    {
+        public override InteractionType InteractionType { get; } = InteractionType.PUSH;
+
+        [Header("Hand Entry")]
+        public double millisecondsCooldownOnEntry;
+        Stopwatch handAppearedCooldown = new Stopwatch();
+
+        [Header("AirPush Detection")]
+        public float speedMin;
+        public float speedMax;
+        public float distAtSpeedMin;
+        public float distAtSpeedMax;
+        public AnimationCurve stiffnessCurve;
+        public float horizontalDecayDist;
+
+        [Header("AirPush Angles")]
+        [Range(0, 180)]
+        public float thetaOne;
+        [Range(0, 180)]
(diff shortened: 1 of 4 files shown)


TF_Service_dotNet/TouchFree/Interactions/InteractionModules/AirPushInteraction.cs
```diff
@@ -1,70 +1,63 @@
...
-        [Range(0, 180)]
-        public float thetaTwo;
+        public float speedMin = 0.15f;
+        public float speedMax = 0.5f;
+        public float distAtSpeedMin = 0.03f;
+        public float distAtSpeedMax = 0.005f;
+        public float horizontalDecayDist = 0.05f;
+
+        public float thetaOne = 15f;
+        public float thetaTwo = 135f;
```
MattGrayUL, line 24: Values like these probably don't need to be public anymore, they were in Unity for convenience of using the inspector to change values, but are not used elsewhere AFAIK

TF_Service_dotNet/TouchFree/Interactions/InteractionModules/AirPushInteraction.cs
```diff
@@ -277,14 +272,14 @@ private float GetAppliedForceChange(float _currentVelocity, float _dt, Vector2 _
                     // Adjust force based on spring stiffness
 
                     // Perform a calculation:
-                    float vClamped = Mathf.Clamp(Mathf.Abs(_currentVelocity), speedMin, speedMax);
+                    float vClamped = Math.Clamp(Math.Abs(_currentVelocity), speedMin, speedMax);
 
                     float ratio = (vClamped - speedMin) / (speedMax - speedMin);
```
MattGrayUL: Can this be more descriptive now that it isn't explicitly used against a 'stiffness curve'? Perhaps stiffnessRatio?

TF_Service_dotNet/TouchFree/Interactions/InteractionModules/AirPushInteraction.cs
NatGreenwood-UL: I'd left these all open as we had talked about adding configuration layers for deeper control of the interactions later down the line, and as the Interactions are part of the _Library_ I can't even mark them as internal without blocking the ability of the Service to chang them.
We _could_ plumb it all through the config system but I wanted to leave this stuff open to avoid _having_ to do it. 

All that said, leaving things in a worse state to support a _theoretical_ feature is a good way to develop technical debt so if you'd rather have things locked down until we need them looser, I'm happy to go back and do that.

MattGrayUL: I don't think there's real harm with leaving it so to aid with customising later, perhaps we leave them as public :)

MattGrayUL (approved):

PR #194, opened 2021-11-29
Corrected the reference to Client to disable it when required

MattGrayUL:


PeterNancollis-UH (approved):
Scene reference looks correct and client enables/disables correctly with focus changes

PR #201, opened 2022-01-17
rotate camera feeds

MattGrayUL:
- [x] Code Review 

### DIFF
--- a/TF_Service_and_Tooling_Unity/Assets/TouchFree/ServiceUI/Prefabs/Camera Previews.prefab
+++ b/TF_Service_and_Tooling_Unity/Assets/TouchFree/ServiceUI/Prefabs/Camera Previews.prefab
@@ -1,5 +1,116 @@
 %YAML 1.1
 %TAG !u! tag:unity3d.com,2011:
+--- !u!1 &172535804162709658
+GameObject:
+  m_ObjectHideFlags: 0
+  m_CorrespondingSourceObject: {fileID: 0}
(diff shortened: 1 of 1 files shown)


TF_Service_and_Tooling_Unity/Assets/TouchFree/ServiceUI/Prefabs/Camera Previews.prefab
```diff
@@ -1,76 +1,107 @@
 %YAML 1.1
 %TAG !u! tag:unity3d.com,2011:
---- !u!1 &308944571178147991
+--- !u!1 &172535804162709658
```
NatGreenwood-UL, line 3: Had a bit of a look over this, the new version of the prefab seems good but there's quite a bit of deleted & added content, not really sure how to easily compare the two. Do you hae a quick rundown of what changed or shall I try and get two copies open for a side-by-side view?

MattGrayUL: Effectively I rotated the two feeds 180, so i'm not sure why there are so many changes. I expect it's because in doing so I applied some changes from the parent prefab (but I discarded the subsequent changes to the parent prefab n the commit) so the parent prefab may be acting as if it is overriding values, but the overrides now match the child

Either way, I simply changed the rotations of the two images

NatGreenwood-UL (approved):

PR #205, opened 2022-01-25
TF-300: TouchFree running on BrightSign

NatGreenwood-UL:
- [x] Code Review 

The contents of the ARMLinux build job artefact should contain everything required to run TouchFree. To actually run TouchFree on a BrightSign box:

1. Copy the contents of the aforementioned artefact onto a microsd card
2. Place that card into a development BrightSign box with `telnet` enabled
3. Connect via `telnet` to that BrightSign box
4. Copy the contents of the card to `/tmp/`
5. Create the following symlinks:
  5a. `libatomic.so.1.2.0 -> libatomic.so.1`
  5b. `libLeapC.so.5 -> libLeapC.so`
  5c. `libstdc++.so.6.0.28 -> libstdc++.so.6`
6. Run the service using this command: `LD_LIBRARY_PATH=. ./TouchFree\ Service`

The above will be covered for production environments as part of the work in TF-302


TF_Service_dotNet/TouchFree/Configuration/ConfigFileUtils.cs
```diff
@@ -25,7 +25,7 @@ public static string ConfigFileDirectory
             }
         }
 
-        static readonly string DefaultConfigDirectory = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.CommonApplicationData), "Ultraleap\\TouchFree\\Configuration\\");
+        static readonly string DefaultConfigDirectory = RuntimeInformation.IsOSPlatform(OSPlatform.Linux) ? Path.GetFullPath("/storage/sd/ultraleap/touchfree/configuration") : Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.CommonApplicationData), "Ultraleap\\TouchFree\\Configuration\\");
```
PeterNancollis-UH: This line is quite long, might be nicer if we split the conditions onto separate lines but not essential

PeterNancollis-UH:
Would it be worth putting the commands needed to run into a small script? Would make life easier for manual testing and help ensure the knowledge of what to do stays with the code. 

Or do you think this will be covered by the next piece of work?

TF_Service_dotNet/TouchFree/Configuration/ConfigFileUtils.cs
NatGreenwood-UL: It was originally multiple lines that I recombined just in case it was the cause of the syntax error being noticed when publishing (as opposed to building). Will revert that back shortly 👍

NatGreenwood-UL:
> Would it be worth putting the commands needed to run into a small script? Would make life easier for manual testing and help ensure the knowledge of what to do stays with the code.
> 
> Or do you think this will be covered by the next piece of work?

TF-302 covers creating the BrightSign extension of the service, which I had assumed was going to be the primary method we use builds on BrightSign boxes (as it does the installation and the launching once set up) so I had left that out of this story. If you think we'll use the build in the form it's in today rather than the extension, I'm happy to add a script to cover it.

PeterNancollis-UH:
I don't expect we will be installing manually beyond that next ticket so was mainly just to make sure the knowledge was shared. Happy to leave it as the instructions in this PR are sufficient and it won't be necessary for long

PeterNancollis-UH (approved):

PR #208, opened 2022-01-28
Added unit tests for positioning module and accompanying refactor

GabrielHomewoodUL:
- [ ] Code Review 


TF_Service_dotNet/TouchFree/VirtualScreen.cs
```diff
@@ -64,7 +64,8 @@ public float DistanceFromScreenPlane(Vector3 worldPosition)
 
             RaycastAgainstPlane(PhysicalScreenPlane, r, out float distanceFromPlane);
 
-            return distanceFromPlane;
+            // Take the absolute value here to normalise against negative values
+            return Math.Abs(distanceFromPlane);
```
GabrielHomewoodUL: I am not sure about this. I think there might be a slight issue with position when calculating the nearest bone. I had a look at the unity code but there isn't the translation with having the z coordinate being negative so it's not entirely clear

TF_Service_dotNet/TouchFreeTests/PositioningModuleTests.cs
```diff
@@ -0,0 +1,124 @@
...
+namespace TouchFreeTests
+{
+    class PositioningModuleTests
+    {
+        PositioningModule sut;
+        TestVirtualScreenManager virtualScreenManager = new TestVirtualScreenManager();
+
+        public PositioningModuleTests()
+        {
+            // TODO: Remove dependency on the Trackers here and instead use vector3 from mocked tracker instead of hand coordinates to make the tests simpler
```
GabrielHomewoodUL: This is a bit of a nice to have, I can put this in as part of this PR, but as it is the two tests for calculate positions will also test the index tip position tracker, I originally used that as it had the most simple hand to position conversion. I could create a test instance and directly return the position without exercising any tracker code instead (what this TODO is referencing)

TF_Service_dotNet/TouchFree/VirtualScreen.cs
MattGrayUL: I think we may still want to allow the return of negative values here? it would signify the user is past the virtual screen plane which:

- could be useful to know in certain circumstances
- would still result in the 'nearest bone being the one furthest from the user (if stood in front of the virtual screen facing it)

TF_Service_dotNet/TouchFreeTests/PositioningModuleTests.cs
GabrielHomewoodUL: I have done this

TF_Service_dotNet/TouchFree/VirtualScreen.cs
GabrielHomewoodUL: I have removed this absolute and resolved an issue with positioning in the code

TF_Service_dotNet/TouchFreeTests/PositionTrackers/NearestTrackerTests.cs
```diff
@@ -0,0 +1,45 @@
...
+            finger.bones = new Leap.Bone[] {
+                new Leap.Bone() { NextJoint = firstJointPosition },
+                new Leap.Bone() { NextJoint = secondJointPosition }
+            };
+
+            //When
+            Vector3 position = sut.GetTrackedPosition(hand);
+
+            //Then
+            Assert.AreEqual(1f, position.X);
```
NatGreenwood-UL: My preference for tests is to have the goal values spelled out explicitly as a variable to make it a little easier to see the relationship between input and output. I'm open to leading on this but worth thinking about now whilst we're early in the process.

TF_Service_dotNet/TouchFreeTests/PositionTrackers/WristTrackerTests.cs
```diff
@@ -0,0 +1,33 @@
...
+    public class WristTrackerTests
+    {
+        WristTracker sut;
+
+        public WristTrackerTests()
+        {
+            sut = new WristTracker();
+        }
+
+        [TestCase(1000, 2000, 3000, 1, 2, 3)]
```
NatGreenwood-UL: Can these only take built-in types or can they take any object? I'm thinking if these are a pair of vector3's it'd be nice to capture them in that form for ease of reading.

TF_Service_dotNet/TouchFreeTests/PositioningModuleTests.cs
```diff
@@ -0,0 +1,126 @@
...
+
+        private IPositionTracker CreateTrackerWithPosition(TrackedPosition trackedPosition, float x, float y, float z)
+        {
+            Mock<IPositionTracker> mockTracker = new Mock<IPositionTracker>();
+            mockTracker.SetupGet(x => x.TrackedPosition).Returns(trackedPosition);
+            mockTracker.Setup(x => x.GetTrackedPosition(It.IsAny<Leap.Hand>())).Returns(new Vector3(x, y, z));
+            return mockTracker.Object;
+        }
+
+        [TestCase(0, 0.5f, 3, 100, 200, 1, 50, 100, 3)]
```
NatGreenwood-UL: Same comment as previous here, if we can wrap these in more complex types (or even just arrays) for readability, I'd prefer that 👍

NatGreenwood-UL (commented):
Few comments about readability, all the refactor stuff looks good. Haven't reviewed the test values themselves but will try out a build manually tomorrow to confirm all works. 👍

TF_Service_dotNet/TouchFree/Interactions/PositionStabiliser.cs
```diff
@@ -20,16 +20,31 @@ public class PositionStabiliser : IPositionStabiliser
...
+
+        public void SetDeadzoneOffset()
+        {
+            deadzoneOffset = previousPositionDeadzoneCurrentSize - lastRawPos;
+        }
+
+        public void ReduceDeadzoneOffset()
+        {
+            deadzoneOffset *= 0.9f;
+        }
```
MattGrayUL: Can we discuss whether adding the deadzone offset should exist as a separate PR or if it is okay to be included here? @PeterNancollis-UH @NatGreenwood-UL

TF_Service_dotNet/TouchFreeTests/PositionTrackers/IndexTipTrackerTests.cs
```diff
@@ -13,12 +13,17 @@ public IndexTipTrackerTests()
...
-        [TestCase(500, 1000, 2000, 0.5f, 1, 2)]
-        public void CalculatePositions_ValidHandPosition_Returns2dVector(float fingerTipX, float fingerTipY, float fingerTipZ, float cursorX, float cursorY, float distanceFromScreen)
+        private static object[] positionsData = new object[]
+        {
+            new object[] { new Vector3(1000, 2000, 3000), new Vector2(1, 2), 3 },
+            new object[] { new Vector3(500, 1000, 2000), new Vector2(0.5f, 1), 2 }
+        };
+
+        [TestCaseSource(nameof(positionsData))]
+        public void CalculatePositions_ValidHandPosition_Returns2dVector(Vector3 wristPositionMm, Vector2 expectedPositionM, float distanceFromScreenM)
```
MattGrayUL, line 23: Not overly keen on the formatting of this, but I don't know of a better way

I think the readability of this is better than before though so I am happy with it :) (particularly as parameters)

TF_Service_dotNet/TouchFree/Interactions/PositionStabiliser.cs
GabrielHomewoodUL: I can back this change out no problem. It's not fully implemented yet either as none of the interactions have been set to use it.

MattGrayUL: If it's not used currently then yeah, I think it best to keep it for a later PR. Happy to approve after that :)

NatGreenwood-UL (approved):
Liking these changes and I think the rework to the Positioning Module -> Virtual Screen relationship looks great as well 👍

PR #209, opened 2022-01-31
Correct air push theta to be in degrees

PeterNancollis-UH:
Fixes a bug spotted in Air Push where the angle of approach (theta) was not being converted to degrees correctly and would have constantly updated even with lateral movements. 

- [x] Code Review
- [x] Developer key item tests


MattGrayUL (approved):
Tested locally and AirPush works as expected, the code makes sense too

PR #210, opened 2022-01-31
Feature/tf 302 CI Generates TouchFree BrightSign extension

NatGreenwood-UL:
- [ ] Code Review 

To use the extension:
* Copy the .tar.gz from the artefact onto a memory card & insert it into the BrightSign box
* `telnet` onto the box, `cp` the archive to `/tmp` and unzip using `tar -xzf ./{archive}`
* run the installation script unpacked by the step above
* the extension should now be installed, reboot the box to mount the extension using `reboot`
* reconnect to the BrightSign box and `cd` to the extension, which should be mounted in `/var/volatile/bsext/ext_UL_TouchFree_Service/`
* run the `./bsext_init` script (still unsure why this isn't being run on boot)
* see TouchFree running in the tasklist by running `top`


Scripts/BrightSign_Scripts/bsext_init
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+PIDFILE=/var/run/haptic-server.pid
```
PeterNancollis-UH: We probably want to rename this pid file to touchfree, we don't have a haptic server any more

.gitlab-ci.yml
```diff
@@ -439,6 +439,32 @@ TF_Service::package:
...
+
+    - cd "$CI_PROJECT_DIR/Scripts/BrightSign_Scripts/"
+    - chmod +x ./make-extension-ubi
+    - apt-get update
+    - apt-get --assume-yes install squashfs-tools
+    - ./make-extension-ubi "$CI_PROJECT_DIR/Service_dotNet_Build/" TF_Service_BS
+  artifacts:
+    name: "TouchFree_Service_Extension"
+    paths:
+      - "$CI_PROJECT_DIR/Scripts/BrightSign_Scripts/build-TF_Service_BS"
```
PeterNancollis-UH: The artefacts end up several directories deep in the zip, could we bring them up to the top level to make it a bit tidier?

PeterNancollis-UH (approved):
Approving even through the extension doesn't start automatically.

The install of the extension works as expected and it can be run manually for testing

PR #212, opened 2022-02-02
TF-317 Combined Tracking / TouchFree BS Extension

NatGreenwood-UL:
- [x] Code Review 

To install the new extension:

* Copy the .tar.gz from the build artefacts onto a MicroSD Card & plug it into a BrightSign box, then restart the box
* Connect to the box via telnet & copy the .tar.gz to the `/tmp/` directory
* Uncompress it with `tar -xzf ./{filename}`
* run the `ext_TF_[...]_install.sh` script to install the extension
* `reboot` the box & reconnect with telnet again (this is a good time to remove the SD card)
* cd to `/var/volatile/bsext/ext_TF[...]`
* run `./bsext_init start` to start both TF and tracking (you can see them running in the `top` task manager)
* Connect the SD card again with the content you want to test, which should now work!


Scripts/BrightSign_Scripts/Download_Tracking.py
```diff
@@ -0,0 +1,54 @@
...
+	os.remove(trackingZip)
+
+if os.path.isdir(trackingUnzippedDir):
+	print("Removing Unzipped Pickler Directory")
+	shutil.rmtree(trackingUnzippedDir)
+
+print("Downloading Artifacts from GitLab")
+
+# Tracking Project ID: 455
+# ARM Release Build Job ID: 485691
```
MattGrayUL, line 36: do these artefacts expire? 

🤷

MattGrayUL (approved):

Scripts/BrightSign_Scripts/Download_Tracking.py
NatGreenwood-UL: The version Alex had _was_ set not to expire but they have removed that now. Unfortunately this means that this build stage cannot work until Alex gets us the GitLab hosted copy of the release version, which I guess means holding off on this merge until then. 😞

NatGreenwood-UL: This has now been switched for an updated build which removes the concerning debug symbols but will persist on our GL. 👍

PR #214, opened 2022-02-04
Feature/TF-318 Quick start on BS

PeterNancollis-UH:
Packages the quick-start example from tooling with an autorun file for testing on Brightsign 

- [x] Code Review
- [x] Developer key Item Tests
- [ ] Product owner review

To test:
* Copy the autorun.brs and "Tooling-Package" directory to the root of a micro SD card
* With TouchFree running, insert the card to load the HTML


Scripts/BrightSign_Scripts/QuickStart-autorun.brs
```diff
@@ -0,0 +1,33 @@
...
+
+Sub Main()
+	x = 0
+	y = 0
+	width = 1920
+	height = 1080
+	rect = CreateObject("roRectangle", x, y, width, height)
+	
+	url = "file:/SD:/Tooling_Package/quick-start/Quick-Start_Example.html"
+
```
MattGrayUL, line 18: This seems to want to render only the corner of the Quick Start example on a 2560x1440 monitor, I have tried:

- width+Height set to 1920x1080
- width+Height set to 2560x1440

I get the same visual result with both settings

MattGrayUL: Setting the resolution in the autorun.brs file to 1280x720 foxes this issue

I suggest we add a comment explaining that as-well-as change the default to it

MattGrayUL (approved):
one suggested change

🙈

Scripts/BrightSign_Scripts/QuickStart-autorun.brs
PeterNancollis-UH: Have added a comment to explain to use 1280x720. 
Have not changed the default as this causes the content to draw in a window on a 1080 display

PR #218, opened 2022-02-10
TF-248: Rationalise units across dot net service

NatGreenwood-UL:
- [ ] Code Review 

Unify measurement units (to mm) across the .net service, with two exceptions:

* Values measured in M within interactions themselves, as these are under change at the moment
* Deadzone sizes, as deadzone is tied with Hover and Hold. I will maybe do this in another pass if we thing it should be done.


MattGrayUL (approved):
Hover and Hold-based changes could be adjusted in a separate piece of work so I agree with leaving them (along with the other interaction settings)

PR #220, opened 2022-02-11
Feature/tf 274 replace unity service with dot net

MattGrayUL:
- [ ] Code Review 


TF_Service_dotNet/TouchFree_Service/Program.cs
```diff
@@ -7,6 +11,14 @@ public class Program
     {
         static void Main(string[] args)
         {
+#if !DEBUG
+            FileStream filestream = new FileStream("log.txt", FileMode.Create);
```
NatGreenwood-UL, line 15: Not sure how feasbile this is when booting the app but it would be nice for this to go into the configured TF Config directory rather than alongside the executable. Not a hard requirement for approval though.

NatGreenwood-UL (approved):
Have reviewed the three sets of commits linked in slack and tested locally, one comment about the logfile location with the .NET Service potentially being more conveniently located but otherwise am good for this to be merged.

PR #222, opened 2022-02-15
Experimental/add tooling library

GabrielHomewoodUL:


TF_Service_dotNet/TouchFree.Tooling/InputActionManager.cs
```diff
@@ -0,0 +1,83 @@
...
+        public static event InputActionEvent TransmitRawInputAction;
+
+        public static InputActionManager Instance;
+
+        // Variable: plugins
+        // A pre-defined plugin array of <ToggleablePlugins> that modify incoming <InputActions>
+        // based on custom rules.
+        ToggleablePlugin[] plugins;
+
+        public void Awake()
```
PeterNancollis-UH: The awake method is a hangover from Unity, and seems to make for an awkward workflow when using this lib directly. Could this be exchanged for a constructor?

TF_Service_dotNet/TouchFree.Tooling/ToggleablePlugin.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace Ultraleap.TouchFree.Tooling
+{
+    // Struct: ToggleablePlugin
+    // A Data structure used to toggle the use of plugins.
+    [System.Serializable]
+    internal struct ToggleablePlugin
```
PeterNancollis-UH: As @MattGrayUL said elsewhere, these toggleable plugins were just a way of providing an easy UI in the Unity version of the tooling. I don't see any value in them in a C# tooling library

PR #223, opened 2022-02-16
Bug/tf 393 tracking mode launch not set

MattGrayUL:
- [ ] Test plan updated
- [x] Code Review 
- [ ] Execution of the updated test plan (changed items for this work only) 

### DIFF
--- a/TF_Service_and_Tooling_Unity/Assets/TouchFree/ServiceShared/Scripts/HandManager.cs
+++ b/TF_Service_and_Tooling_Unity/Assets/TouchFree/ServiceShared/Scripts/HandManager.cs
@@ -111,17 +111,13 @@ namespace Ultraleap.TouchFree.ServiceShared
                 yield return null;
             }
 
-            Debug.Log("got leap controller");
-
+            // Ensure the leap controller has its tracking mode set after it connects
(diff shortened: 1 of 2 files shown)


TF_Service_and_Tooling_Unity/Assets/TouchFree/ServiceShared/Scripts/HandManager.cs
```diff
@@ -109,6 +110,13 @@ IEnumerator UpdateTrackingAfterLeapInit()
                 yield return null;
             }
 
+            // Ensure the leap controller has its tracking mode set after it connects
+            trackingProvider.GetLeapController().Connect += TrackingConnected;
+            UpdateTrackingTransformAndMode();
```
PeterNancollis-UH, line 115: Is this immediate call to set the transform required?

MattGrayUL: It is, this is the old way that we would set the tracking mode


I didn't actually edit that, the PR has shown the wrong line as new because they are the same

The line in the box is actually the new one:
image

PeterNancollis-UH (approved):
Code change looks good, and I can't recreate the issue with this version installed

PR #225, opened 2022-02-16
Bug/fix meter to milimeter conversion

GabrielHomewoodUL:
- [ ] Code Review 

### DIFF
--- a/TF_Service_dotNet/TouchFree/Interactions/PositionStabiliser.cs
+++ b/TF_Service_dotNet/TouchFree/Interactions/PositionStabiliser.cs
@@ -81,10 +81,9 @@ namespace Ultraleap.TouchFree.Library.Interactions
             return constrainedPositionCurrent;
         }
 
-        public Vector2 ApplyDeadzoneSized(Vector2 previous, Vector2 current, float radiusM)
+        public Vector2 ApplyDeadzoneSized(Vector2 previous, Vector2 current, float radiusMm)
         {
             Vector2 constrainedPosition;
-            float radiusMm = radiusM * 1000f;
 
             float distance = Vector2.Distance(previous, current);
             if (distance > radiusMm)
@@ -132,8 +131,7 @@ namespace Ultraleap.TouchFree.Library.Interactions
 
             if (previousConstraintVector != Vector2.Zero)
             {
-                float distanceAwayFromConstraint = Vector2.Dot(defaultPositionChange, previousConstraintVector) / previousConstraintVector.Length();
-                distanceAwayFromConstraint = Math.Max(0, distanceAwayFromConstraint) / 1000;
+                float distanceAwayFromConstraint = Math.Abs(Vector2.Dot(defaultPositionChange, previousConstraintVector) / previousConstraintVector.Length());
 
                 float shrinkDistance = distanceAwayFromConstraint * shrinkingSpeed;
                 currentDeadzoneRadius -= shrinkDistance;
@@ -157,7 +155,7 @@ namespace Ultraleap.TouchFree.Library.Interactions
 
         void OnSettingsUpdated(InteractionConfig _config)
         {
-            defaultDeadzoneRadius = _config.DeadzoneRadius;
+            defaultDeadzoneRadius = _config.DeadzoneRadius * 1000;
         }
     }
 }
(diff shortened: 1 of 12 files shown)


TF_Service_dotNet/TouchFree/Interactions/PositionStabiliser.cs
```diff
@@ -132,8 +131,7 @@ void ShrinkDeadzone(Vector2 constrainedPositionDefault)
 
             if (previousConstraintVector != Vector2.Zero)
             {
-                float distanceAwayFromConstraint = Vector2.Dot(defaultPositionChange, previousConstraintVector) / previousConstraintVector.Length();
-                distanceAwayFromConstraint = Math.Max(0, distanceAwayFromConstraint) / 1000;
+                float distanceAwayFromConstraint = Math.Abs(Vector2.Dot(defaultPositionChange, previousConstraintVector) / previousConstraintVector.Length());
```
NatGreenwood-UL: Could the difference between Max and Abs here cause issues with shrinking deadzones if they do stray into a negative value?

TF_Service_dotNet/TouchFreeTests/VirtualScreenTests.cs
```diff
@@ -41,29 +41,29 @@ public void Constructor_ValidInputs_ReturnsInstance()
 
         private int ScreenWidthInPixels = 1080;
         private int ScreenHeightInPixels = 1920;
-        private float ScreenHeightInMeters = 400f;
+        private float ScreenHeightInMilimeters = 400f;
```
NatGreenwood-UL: Millimeters*, couple instances of this in this file

NatGreenwood-UL (commented):
Have looked over the code, couple minor comments, but haven't yet tested, going to look to do that now 👍

NatGreenwood-UL:
Having tried this out a bit locally, I think AirPush and TouchPlane are working as expected but H&H feels _very_ sticky, I would guess there's some unit confusion going on with the Deadzone scaling that H&H uses maybe? Feels hard _not_ to start a click on my machine, with the responsiveness set to the same as that of the other interactions.

Also encountered some confusion with TouchPlane and the Interaction Zone; turning on the Interaction Zone with its default distances (which gives a range of 0cm to 25cm from the screen) doesn't seem to allow _dragging_ with TouchPlane, even thought the default plane distance is 5cm from the screen? One or both may be being incorrectly offset

TF_Service_dotNet/TouchFree/Interactions/PositionStabiliser.cs
GabrielHomewoodUL: The Math.Abs will make it so that this is always positive. Should we be ignoring the shrinking if the distance is negative?

TF_Service_dotNet/TouchFreeTests/VirtualScreenTests.cs
GabrielHomewoodUL: I've updated the spelling on these now

TF_Service_dotNet/TouchFree/Interactions/PositionStabiliser.cs
NatGreenwood-UL: I'd guess that if we're trying to shrink the Deadzone by a negative amount (I guess _growing_ the Deadzone) that we'd want to do nothing rather than continue to shrink it regardless?

Equally I don't think we want to allow growing the deadzone by these means as there is a mechanism to stop the shrinking once it hits a threshold that wouldn't be hit going larger.

TF_Service_dotNet/TouchFree/Interactions/PositionStabiliser.cs
```diff
@@ -157,7 +155,7 @@ public void ScaleDeadzoneByProgress(float _progressToClick, float _maxDeadzoneIn
 
         void OnSettingsUpdated(InteractionConfig _config)
         {
-            defaultDeadzoneRadius = _config.DeadzoneRadius;
+            defaultDeadzoneRadius = _config.DeadzoneRadius * 1000;
```
NatGreenwood-UL: As this is the only use of the Deadzone from the config, could this unit change be moved to the loading in `InteractionConfig.cs`?

NatGreenwood-UL (approved):
Tested the latest version and feels good with all interactions 👍

PR #227, opened 2022-02-17
TF-340: TouchFree installs Tracking

NatGreenwood-UL:
- [x] Test plan updated
- [x] Code Review 
- [x] QA Review


Scripts/Download_Tracking.py
```diff
@@ -0,0 +1,54 @@
...
+	shutil.rmtree(trackingUnzippedDir)
+
+print("Downloading Artifacts from GitLab")
+
+# Tracking Project ID: 455
+# 5.4 Windows Production Release Job ID: 501756
+
+TrackingProject = gl.projects.get(455)
+
+ArtefactJob = TrackingProject.jobs.get(501756)
```
GabrielHomewoodUL: Would it be worth having this Job ID parameterised? Just so that when we update to the next version it's easier to update

NatGreenwood-UL: Threw this change in to shift this jobID to our CI file so it's easier to change in future

GabrielHomewoodUL (approved):

PR #228, opened 2022-02-23
Further conditional logging to file

GabrielHomewoodUL:
- [ ] Code Review 


NatGreenwood-UL (commented):
All the code changes look good however the build has failed, as the integration test that checks for the content of the logfile is now looking in the wrong place. Shouldn't be too hard to rectify, it's handled in `Testing\ServiceSocketTests\gulpfile.js`, I think you'll just have to change the default logfile path 👍

NatGreenwood-UL (approved):
Tested this locally on a BrightSign box, unfortunately it doesn't seem to have solved the startup issue but I have confirmed that the file redirect works on the box!

PR #229, opened 2022-02-24
Feature/tf 361 allow query of state

GabrielHomewoodUL:
- [x] Code Review 

### DIFF
--- a/TF_Service_dotNet/TouchFree_Service/Connection/ClientConnection.cs
+++ b/TF_Service_dotNet/TouchFree_Service/Connection/ClientConnection.cs
@@ -114,7 +114,7 @@ namespace Ultraleap.TouchFree.Service.Connection
         public void SendStatus(ServiceStatus _status)
         {
             CommunicationWrapper<ServiceStatus> message =
-                new CommunicationWrapper<ServiceStatus>(ActionCode.STATUS.ToString(), _status);
+                new CommunicationWrapper<ServiceStatus>(ActionCode.SERVICE_STATUS.ToString(), _status);
 
             string jsonMessage = JsonConvert.SerializeObject(message);
 
@@ -178,6 +178,9 @@ namespace Ultraleap.TouchFree.Service.Connection
             {
                 case ActionCode.SET_CONFIGURATION_STATE:
                     receiver.configChangeQueue.Enqueue(content);
+                    break;
+                case ActionCode.REQUEST_SERVICE_STATUS:
+                    receiver.requestServiceStatusQueue.Enqueue(content);
                     break;
                 case ActionCode.REQUEST_CONFIGURATION_STATE:
                     receiver.configStateRequestQueue.Enqueue(content);
--- a/TF_Service_dotNet/TouchFree_Service/Connection/ClientConnection.cs
+++ b/TF_Service_dotNet/TouchFree_Service/Connection/ClientConnection.cs
@@ -145,7 +145,7 @@ namespace Ultraleap.TouchFree.Service.Connection
             }
             else if (clientVersionParsed.Minor > _coreVersion.Minor)
             {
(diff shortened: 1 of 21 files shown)


TF_Service_dotNet/TouchFree_Service/Connection/ClientConnection.cs
```diff
@@ -121,16 +141,16 @@ private Compatibility GetVersionCompability(string _clientVersion, Version _core
 
             else if (clientVersionParsed.Minor < _coreVersion.Minor)
             {
-                return Compatibility.CLIENT_OUTDATED;
+                return Compatibility.CLIENT_OUTDATED_WARNING;
```
NatGreenwood-UL, line 144: Why are we warning incompatibility with Minor versions now? Historically we've said that Minor changes in the API are breaking and only Hotfix changes were compatible?

NatGreenwood-UL (commented):
Just one comment on the code, going to test this now 👍

TF_Service_dotNet/TouchFree_Service/Connection/TouchFreeServiceTypes.cs
```diff
@@ -25,7 +28,28 @@ internal enum Compatibility
     {
         COMPATIBLE,
         SERVICE_OUTDATED,
-        CLIENT_OUTDATED
+        CLIENT_OUTDATED,
+        SERVICE_OUTDATED_WARNING,
+        CLIENT_OUTDATED_WARNING
+    }
+
+    public enum CameraState
```
NatGreenwood-UL: Was thinking about this, currently if the tracking Service isn't connected, we'll report Camera NOT_CONNECTED even if it is physically, as we can't get at the information about the hardware without the Service. Was going to suggest a third entry here along the lines of "NO_SERVICE" to indicate that but then I realised at that point we may as well combine this into one enum, as we can only (reliably) report 

* Everything fine
* No Camera
* No Service

Do you think it's worth the consolidation here as the scenario of "No Service, Camera" will never actually be possible?

GabrielHomewoodUL: I think that makes sense. I'll start making the changes now

TF_Tooling_Web/src/TouchFreeToolingTypes.ts
```diff
@@ -104,19 +104,13 @@ export enum InteractionType {
...
-// CONNECTED - There is a camera connected
-export enum CameraState {
-    NOT_CONNECTED,
-    CONNECTED,
-}
-
 // Enum: TrackingServiceState
-// UNAVAILABLE - The TouchFree services is not connected to the tracking service
+// UNAVAILABLE - The TouchFree service is not connected to the tracking service
+// NOT_CONNECTED - The TouchFree service is connected to the tracking service but there is not a camera connected
```
MattGrayUL: this should be NO_CAMERA :)

GabrielHomewoodUL: Done!

TF_Service_dotNet/TouchFree_Service/Connection/TouchFreeServiceTypes.cs
GabrielHomewoodUL: Done!

TF_Service_dotNet/TouchFree_Service/Connection/ClientConnection.cs
GabrielHomewoodUL: We have come to the decision to allow the service to be minor versions ahead and the client to be hotfix versions ahead. I've pushed up the change for this

MattGrayUL (dismissed):

MattGrayUL (approved):

PR #232, opened 2022-03-02
TF-358: Welcome Screen for Web-based settings UI

NatGreenwood-UL:
# Checklist:
- [x] Test plan updated
- [x] Code Review 
- [x] QA Review

# Testing Procedures:
* Install the build of TouchFree, a build of tracking, and connect a camera
* Navigate to http://localhost:9739/settings/index.html
* Observe the Welcome Screen
* Observe the status icons in the top left (expect both dots to be green)
* Disconnect the camera, and wait 5 seconds
    * The dot for the camera should go red
* Open "Services" in windows, find and stop the Ultraleap Tracking Service
    * The dot for the Service (the hand icon) should go red
* Reconnect the camera, then restart the Tracking Service
    * Both dots should go green again


TF_Service_dotNet/TouchFree_Service/Startup.cs
```diff
@@ -30,12 +30,34 @@ public void ConfigureServices(IServiceCollection services)
...
+            services.AddControllers();
         }
 
         // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
         public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
         {
             app.UseWebSockets();
+            app.UseRouting();
+
+            app.UseStaticFiles("/settings");
```
GabrielHomewoodUL, line 39: I think the UseStaticFiles is the only change you need in this file. I think the rest are just for setting up a Web API

TF_Settings_Web/public/manifest.json
```diff
@@ -0,0 +1,25 @@
...
+      "sizes": "64x64 32x32 24x24 16x16",
+      "type": "image/x-icon"
+    },
+    {
+      "src": "logo192.png",
+      "type": "image/png",
+      "sizes": "192x192"
+    },
+    {
+      "src": "logo512.png",
```
GabrielHomewoodUL: I don't think we want the react logos?

TF_Service_dotNet/TouchFree_Service/Startup.cs
NatGreenwood-UL: I'm pretty sure that we need the `UseCors` because that's the policy with which the files are provided to the browser (If we don't use CORS policies the browser may realise the files are being loaded from the filesystem which they then block). I'll see what I can strip out RE: the others.

NatGreenwood-UL: I take it back, seems like CORS issues aren't present with this hosting method, have stripped this down to just the `UseStaticFiles` without issue.

NatGreenwood-UL:
⚠️ **Note, we should hold off on merging this until [TF-361](https://github.com/ultraleap/TouchFree/pull/229) is merged, as this branch features all of its changes**

TF_Settings_Web/src/Components/ScreenManager.tsx
```diff
@@ -0,0 +1,112 @@
...
+    private inputSystem: any;
+    private connectionManager: any;
+
+    constructor(props: {}) {
+        super(props);
+
+        this.timerID = -1;
+
+        TouchFree.Connection.ConnectionManager.init();
+        this.inputSystem = new TouchFree.InputControllers.WebInputController();
```
GabrielHomewoodUL, line 46: Can we import the classes from tooling to make them a bit easier to use? e.g. `import { WebInputController } from "../TouchFree/TouchFree_Tooling"`

GabrielHomewoodUL (approved):

TF_Settings_Web/src/Components/ScreenManager.tsx
NatGreenwood-UL: I tried to import them using this method and had little to no success getting it to actually work in-browser locally as they would always cause the loading of this file to fail entirely :/

PR #235, opened 2022-03-07
Ensure logfile directory is created before trying to use it

NatGreenwood-UL:
This resolves an issue with starting the TouchFree Service on BS, enabling me to test 361 on the box (which was successful).

- [ ] Code Review 
- [ ] QA Review

### DIFF
--- a/TF_Service_dotNet/TouchFree_Service/Program.cs
+++ b/TF_Service_dotNet/TouchFree_Service/Program.cs
@@ -16,7 +16,9 @@ namespace Ultraleap.TouchFree.Service
 #if !DEBUG
             var loggingFileDirectory = RuntimeInformation.IsOSPlatform(OSPlatform.Linux) ? ConfigFileUtils.ConfigFileDirectory : "";
 
-            Directory.CreateDirectory(loggingFileDirectory);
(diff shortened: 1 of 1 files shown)


TF_Service_dotNet/TouchFree_Service/Program.cs
```diff
@@ -15,6 +15,11 @@ static void Main(string[] args)
         {
 #if !DEBUG
             var loggingFileDirectory = RuntimeInformation.IsOSPlatform(OSPlatform.Linux) ? ConfigFileUtils.ConfigFileDirectory : "";
+
+            if (loggingFileDirectory != "") {
+                Directory.CreateDirectory(loggingFileDirectory);
```
GabrielHomewoodUL, line 20: Do we need to check whether the folder already exists or does this handle that? I can check but just wondering if you'd already tested it

NatGreenwood-UL: I ended up putting the check in because it failed the integration tests, as you can't give CreateDirectory an empty string. 👍

GabrielHomewoodUL (approved):

PR #237, opened 2022-03-10
TF-400: API for saving configs via WS

NatGreenwood-UL:
- [x] Test plan updated
- [x] Code Review 
- [x] QA Review


TF_Application/Assets/TouchFree/Tooling/Scripts/Connection/TouchFreeServiceTypes.cs
```diff
@@ -127,6 +188,7 @@ public struct ConfigChangeRequest
         public ConfigChangeRequest(string _id)
         {
             requestID = _id;
+            requestID = _id;
```
GabrielHomewoodUL: Duplicate line here

TF_Settings_and_Tooling_Unity/Assets/TouchFree/Tooling/Scripts/Connection/TouchFreeServiceTypes.cs
```diff
@@ -179,6 +188,7 @@ public struct ConfigChangeRequest
         public ConfigChangeRequest(string _id)
         {
             requestID = _id;
+            requestID = _id;
```
GabrielHomewoodUL: Another duplicated line

GabrielHomewoodUL (approved):

TF_Settings_and_Tooling_Unity/Assets/TouchFree/Tooling/Scripts/Connection/TouchFreeServiceTypes.cs
NatGreenwood-UL: Same duplicated line in this case, the version in TouchFree itself is just the Tooling imported from here. Have removed both.

Speaking of, I've been thinking that we should add some form of automation to update TouchFree Overlay's Tooling instance. Many of the "changes" to the Tooling used in TouchFree itself are due to the Tooling having not been updated. Would be good to have TouchFree reliably using the latest, as it might help us find issues with the tooling, and avoids a scenario where the Service updates the API version but the Tooling in TouchFree isn't updated to match.

PR #241, opened 2022-03-16
Feature/tf 436 add tracking connection retry

GabrielHomewoodUL:
- [x] Code Review 

### DIFF
--- a/TF_Service_dotNet/TouchFree/TrackingConnectionManager.cs
+++ b/TF_Service_dotNet/TouchFree/TrackingConnectionManager.cs
@@ -43,7 +43,6 @@ namespace Ultraleap.TouchFree.Library
 
                 if (!controller.IsServiceConnected)
                 {
-                    Console.WriteLine($"Attempting reconnect after {waitTimeSeconds} seconds");
                     controller.StartConnection();
                 }
                 else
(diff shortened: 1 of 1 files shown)


TF_Service_dotNet/TouchFree/TrackingConnectionManager.cs
```diff
@@ -8,20 +9,52 @@ public class TrackingConnectionManager
...
+
         public TrackingConnectionManager(IConfigManager _configManager)
         {
             configManager = _configManager;
             controller = new Leap.Controller();
             controller.Connect += Controller_Connect;
+            controller.Disconnect += Controller_Disconnect;
             UpdateTrackingMode(_configManager.PhysicalConfig);
             _configManager.OnPhysicalConfigUpdated += UpdateTrackingMode;
+            CheckConnectionAndRetryOnFailure();
```
MattGrayUL, line 24: Is this required when `controller = new Leap.Controller();` exists above? will it not immediately be broken?

GabrielHomewoodUL: This is to handle the situation where the tracking service isn't running when the touchfree service starts. Instantising the Controller makes it attempt to connect, but it won't retry if the tracking service isn't available.

MattGrayUL (approved):

PR #244, opened 2022-03-18
TF-360: Interaction Config screen in web settings UI

NatGreenwood-UL:
- [x] Code Review 
- [x] QA Review

To access the settings UI, with a build of TouchFree Service from this branch active, access `http://localhost:9739/settings/index.html`
An autorun.brs for BrightSign testing can also be provided


TF_Settings_Web/src/Components/Pages/InteractionsPage.tsx
```diff
@@ -0,0 +1,307 @@
...
+                        <video autoPlay loop key={this.state.interactionConfig.InteractionType} className="InteractionPreview">
+                            <source src={this.videoPaths[activeInteraction]} type="video/webm" />
+                        </video>
+                    </div>
+
+                    <div className="verticalContainer sideSpacing">
+                        <Slider name="Cursor Movement"
+                            increment={0.001}
+                            rangeMin={0}
+                            rangeMax={1}
```
PeterNancollis-UH: I think this rangeMax value is setting a deadzone in meters where it should be centimeters. With the slider at half way I have to move my hand about 0.5m to make the cursor move

TF_Settings_Web/src/Components/Pages/InteractionsPage.tsx
```diff
@@ -0,0 +1,307 @@
...
+        newConf.InteractionMaxDistanceCm = parseFloat(e.currentTarget.value);
+
+        this.setState(() => ({
+            interactionConfig: newConf
+        }));
+    }
+
+    resetToDefaults(): void {
+        // ???????
+        // How to get an "Defaults?"
```
PeterNancollis-UH: Is it intentional that this isn't implemented?

TF_Settings_Web/src/App.css
```diff
@@ -0,0 +1,30 @@
+.App {
```
PeterNancollis-UH: Not sure the correct place to comment on this, but the app doesn't fill the vertical height of the browser. I think it should fill with the background colour to the full height even when there's only content at the top
IUgS2pkyJF

TF_Settings_Web/src/Components/Pages/InteractionsPage.tsx
NatGreenwood-UL: This is now implemented

TF_Settings_Web/src/Components/Pages/InteractionsPage.tsx
NatGreenwood-UL: This is matched to the range of the slider in Unity now 👌

TF_Settings_Web/src/TouchFree/Configuration/ConfigurationManager.ts
```diff
@@ -0,0 +1,101 @@
+import {
```
PeterNancollis-UH, line 1: Is this a copy of the source for tooling to add it to the settings UI? 

I know we've talked about automating the maintaining of this but just wanted to check

TF_Tooling_Web/.gitignore
```diff
@@ -8,4 +8,4 @@
 ./test/**/*.js
 dist
 
-./Plugins/**/SnappingPlugin.js
+./Plugins/**/Snapping_Plugin.js
```
PeterNancollis-UH, line 11: Why has the snapping plugin been added to gitignore? Is this just removing a locally built version?

NatGreenwood-UL: Yes, the .js form is the built version. The change here is from an old name to a new name, it was always meant to be ignored.

TF_Settings_Web/src/TouchFree/Configuration/ConfigurationManager.ts
NatGreenwood-UL: Yes, the `src/TouchFree` directory is a clone of the Web Tooling for now (until we do the work you mentioned). I've made sure that the two are identical as of the last commit I made but this is definitely another motivator to _either_ do the automatic import, to work out what the workflow for people who want to consume tooling as TypeScript looks like, or possibly both?

PeterNancollis-UH: I think both are important going forward. This is fine for the time being though

TF_Settings_Web/src/Components/Controls/Slider.tsx
```diff
@@ -0,0 +1,87 @@
...
+                <div className="sliderContainer">
+                    <input type="range"
+                        step={this.props.increment}
+                        min={this.props.rangeMin}
+                        max={this.props.rangeMax}
+                        className="slider"
+                        onChange={this.onChange}
+                        onPointerMove={this.onMove.bind(this)}
+                        onPointerDown={this.onDown.bind(this)}
+                        onPointerUp={this.onUpOut.bind(this)}
```
PeterNancollis-UH: I think we need to add an `onPointerLeave` handler calling the `onUpOut` method to prevent the handle getting stuck to the cursor

PeterNancollis-UH (approved):
Tested on BrightSign. All controls can be operated by mouse and TouchFree and although there are a few improvements we still want to make we can deal with those as separate work as this does provide a usable way to configure the interactions as is

PR #250, opened 2022-03-23
TF-435 Scaled cursor position with device pixel ratio

GabrielHomewoodUL:
- [x] Code Review 

### DIFF
--- a/TF_Tooling_Web/src/TouchFreeToolingTypes.ts
+++ b/TF_Tooling_Web/src/TouchFreeToolingTypes.ts
@@ -51,7 +51,7 @@ export class TouchFreeInputAction {
 // Used to translate the raw actions that come across the websocket (<WebsocketInputActions>) and
 // convert them into the Tooling-friendly <TouchFreeInputAction> format.
 export function ConvertInputAction(_wsInput: WebsocketInputAction): TouchFreeInputAction {
-    const yPosition = (window.innerHeight - _wsInput.CursorPosition.y) / window.devicePixelRatio;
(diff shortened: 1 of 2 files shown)


TF_Tooling_Web/src/TouchFreeToolingTypes.ts
```diff
@@ -51,15 +51,16 @@ export class TouchFreeInputAction {
 // Used to translate the raw actions that come across the websocket (<WebsocketInputActions>) and
 // convert them into the Tooling-friendly <TouchFreeInputAction> format.
 export function ConvertInputAction(_wsInput: WebsocketInputAction): TouchFreeInputAction {
-    const yPosition = window.innerHeight - _wsInput.CursorPosition.y;
+    const yPosition = (window.innerHeight - _wsInput.CursorPosition.y) / window.devicePixelRatio;
```
PeterNancollis-UH: This is doing the operations in the wrong order, so the cursor ends up offset as we subtract the unscaled window height

GabrielHomewoodUL: I have updated this!

PeterNancollis-UH: Working great after the fix, the y position of the cursor is consistent at different scalings

PeterNancollis-UH:
A knock of effect of changing the position scaling is that the cursor animation is now misaligned when the screen is scaled, so it shrinks towards the bottom corner rather than the middle

chrome_7A9SXD70cW

NatGreenwood-UL (approved):
I've confirmed that the offset experienced by Pete above is no longer present and am happy with the code 👍

PR #251, opened 2022-03-25
Feature/tf 359 camera setup screen in web settings UI

GabrielHomewoodUL:
- [x] Code Review 


TF_Settings_Web/src/Components/Controls/TextEntry.tsx
```diff
@@ -0,0 +1,29 @@
...
+    selected: boolean,
+    onChange: (ChangeEventHandler<HTMLInputElement>),
+    onClick: (MouseEventHandler<HTMLElement>),
+    onPointerDown: (PointerEventHandler<HTMLElement>)
+}
+
+export class TextEntry extends React.Component<TextEntryProps, {}> {
+    render() {
+        return(
+            <label onClick={this.props.onClick.bind(this)} onPointerDown={this.props.onPointerDown.bind(this)} className={"textEntryBackgroundLabel " + (this.props.selected ? "textEntryBackgroundLabelSelected" : "")}>
```
NatGreenwood-UL: Could we move the logic in classname into the body of the `render` function and split this line by each prop? It's a little hard to read at this length.

TF_Settings_Web/src/Components/Controls/TextEntry.tsx
```diff
@@ -0,0 +1,29 @@
...
+                <p className="textEntryLabel">{this.props.name}</p>
+                <label className="textEntryContainer">
+                    <input type="number"
+                           step={0.01}
```
NatGreenwood-UL: Do we need the `step` property here? We're hiding the scroll buttons so there's no way to increment this item (I realise the same may be true of the number input in TextSlider in retrospect)

TF_Settings_Web/src/Components/Pages/CameraPage.tsx
```diff
@@ -1,18 +1,240 @@
...
+        };
+
+        this.state = state;
+    }
+
+    componentDidMount(): void {
+        if (ConnectionManager.serviceConnection()?.webSocket?.readyState === WebSocket.OPEN) {
+            ConfigurationManager.RequestConfigFileState(this.setStateFromFile.bind(this));
+        } else {
+            ConnectionManager.instance.addEventListener('OnConnected', () => {
```
NatGreenwood-UL: ConnectionManager has a custom `AddConnectionListener` function which could replace the logic of this function (if already connected, the callback is immediately executed, if not, it's added to a list to be invoked once a connection is made)

TF_Settings_Web/src/Components/Pages/CameraPage.tsx
```diff
@@ -1,18 +1,240 @@
...
+            ConfigurationManager.RequestConfigFileChange(null, physicalConfigState, this.configChangeCbHandler.bind(this));
+        }
+    }
+
+    configChangeCbHandler(result: WebSocketResponse): void {
+        if (result.status !== "Success") {
+            console.error(`Failed to set config state! Info: ${result.message}`);
+        }
+    }
+
```
NatGreenwood-UL, line 115: Got some odd whitespace in here, there's two spaces here but then 0 spaces between a bunch of the functions below?

NatGreenwood-UL (commented):
A few minor comments, the ConnectionManager thing is the most important I think

TF_Settings_Web/src/Components/Pages/CameraPage.tsx
```diff
@@ -1,18 +1,240 @@
...
 
-export class CameraPage extends Page {
-    private divStyle: CSSProperties = {
-        flex: '1 1 auto',
-        width: '100%',
-        height: '800px',
-        backgroundColor: 'red',
-    };
+    onScreenHeightClicked(): void {
+        this.setState({...this.state, selectedView: "screenHeight"});
```
NatGreenwood-UL: Two things here: 

1. there's some inconsistency in the use of an object and lambda inside this.setState to change the values, and I'm not sure there's a reason for it? Would be nice to be consistent if not.
2. We're using `this.state` inside of setState, but setState is executed asynchronously. We should probably be using the `state` parameter that gets passed to the lambda within `setState` to be as up-to-date as possible if we're modifying said state, as per [the docs here](https://reactjs.org/docs/react-component.html#setstate)

GabrielHomewoodUL: I've updated this to only update the selectedView state item. setState handles the properties that are already there that you aren't changing so the use of this.state was unnecessary

NatGreenwood-UL: What about the uses that precede this one where they're using `this.setState(() => ({` ?

GabrielHomewoodUL: I have now updated those as well

NatGreenwood-UL (approved):

NatGreenwood-UL (approved):

PeterNancollis-UH:
Some observations from testing:

- You can't type a negative number in the text fields. You can enter a '-' after a number is in the field, but not before
- The angle of the camera in the lower image shows the camera facing the screen. It should face away so it matches screen top
- The cursor ring is really hard to see (making it hard to test changes to the camera setup) could we change the colour to a light grey / white

PeterNancollis-UH:
☝️ The above have been resolved

PeterNancollis-UH:
During testing I found that if I set the camera to desktop orientation the tracking mode is not set correctly

PR #253, opened 2022-03-28
Release/bs extension alpha1

PeterNancollis-UH:
- [ ] Code review
- [ ] Developer Testing


Scripts/BrightSign_Scripts/Version.txt
```diff
@@ -0,0 +1,2 @@
+TouchFree Extension Version: 1.0.0-alpha1
+TouchFree Service Version: 2.3.0-alpha1
```
NatGreenwood-UL: Should this not also include the API version so that people can align their Tooling versions?

PeterNancollis-UH: Yes, good point. I have updated

PR #254, opened 2022-03-30
TF-420 TF Service - Handle the screen resolution being set with bad values

GabrielHomewoodUL:
- [ ] Test plan updated
- [ ] Code Review 

### DIFF
--- a/TF_Service_dotNet/TouchFree_Service/Connection/ClientConnection.cs
+++ b/TF_Service_dotNet/TouchFree_Service/Connection/ClientConnection.cs
@@ -9,6 +9,7 @@ using System.Text;
 
 using Ultraleap.TouchFree.Library;
 using Ultraleap.TouchFree.Service.ConnectionTypes;
+using Ultraleap.TouchFree.Library.Configuration;
 
 namespace Ultraleap.TouchFree.Service.Connection
 {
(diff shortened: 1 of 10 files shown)


TF_Service_dotNet/TouchFree_Service/Connection/ClientConnection.cs
```diff
@@ -269,38 +272,45 @@ protected void ProcessHandshake(ActionCode action, string requestContent)
             string clientApiVersion = (string)contentObj[VersionInfo.API_HEADER_NAME];
             Compatibility compatibility = GetVersionCompability(clientApiVersion, VersionInfo.ApiVersion);
 
+            string configurationWarning = string.Empty;
+
+            if (!configManager.AreConfigsInGoodState())
+            {
+                configurationWarning = " Configuration is in a bad state. Please update the settings via the Settings UI";
```
MattGrayUL: This should probably be named 'TouchFree Settings' rather than 'Settings UI'

TF_Service_dotNet/TouchFree_Service/Connection/ClientConnection.cs
```diff
@@ -269,38 +272,45 @@ protected void ProcessHandshake(ActionCode action, string requestContent)
...
+            }
+
             switch (compatibility)
             {
                 case Compatibility.COMPATIBLE:
                     HandshakeCompleted = true;
                     response.status = "Success";
-                    response.message = "Handshake Successful";
+                    response.message = "Handshake Successful." + configurationWarning;
```
MattGrayUL, line 287: Should each of these response messages include a space between the string and the warning?

e.g. "Handshake Successful. " + configurationWarning;

GabrielHomewoodUL: I've included the space where the warning is set so that it only has the space when there is a warning

TF_Service_dotNet/TouchFree_Service/Connection/ClientConnection.cs
GabrielHomewoodUL: I've updated this message

MattGrayUL (approved):

PR #262, opened 2022-04-06
Feature/tf 440 simplify tooling consumption in repo

GabrielHomewoodUL:
- [x] Code Review 


TF_Tooling_Web/src/Connection/MessageReceiver.ts
```diff
@@ -131,7 +131,6 @@ export class MessageReceiver {
         if (this.responseCallbacks !== undefined) {
             for (let key in this.responseCallbacks) {
                 if (key === _response.requestID) {
-                    console.log(_response.message);
```
GabrielHomewoodUL, line 134: This was added so that the tooling consumer would get a warning if their tooling version did not match, or if their config was in a bad state

NatGreenwood-UL: I see, in that case I need to completely invert this change and *add* it to the MessageReciever in the Settings UI as they were out of sync (which is why I removed this)

Scripts/setup_repository.bat
```diff
@@ -0,0 +1,10 @@
+@echo off
+if not "%1"=="1" (
+    powershell -Command "Start-Process -Verb RunAs -FilePath '%0' -ArgumentList '1'"
+    exit /b
+)
+
+cd %~dp0
+
+mkdir "../TF_Application/Assets/TouchFree/"
+mklink /J "../TF_Application/Assets/TouchFree/Tooling" "../TF_Settings_and_Tooling_Unity/Assets/TouchFree/Tooling"
```
MattGrayUL, line 10: Tested this .bat on my machine, so I still had some hanging folder meta files in Unity and that meant the .bat didn't copy the files (I assume mkdir errored as there already was one?)

I deleted the folders inside of Unity and then re-ran the .bat and it worked....

I am happy to mark this as known behaviour, but thought it should be noted.

MattGrayUL (approved):

Scripts/setup_repository.bat
NatGreenwood-UL: I'll throw something for this into the ReadMe just for ease 👍

MattGrayUL (approved):

PR #265, opened 2022-04-07
TF-472 - Unity Tooling - Reconnect to service if disconnected, and re-attempt connection if service is not started

GabrielHomewoodUL:
- [x] Code Review 

### DIFF
--- a/TF_Application/Assets/TouchFree/Tooling/Scripts/Connection/ConnectionManager.cs
+++ b/TF_Application/Assets/TouchFree/Tooling/Scripts/Connection/ConnectionManager.cs
@@ -55,7 +55,7 @@ namespace Ultraleap.TouchFree.Tooling.Connection
         // This value is settable in the Inspector.
         [SerializeField] string port = "9739";
 
-        static bool disconnected = true;
+        static bool shouldReconnect = false;
 
         // Group: Functions
 
@@ -77,7 +77,7 @@ namespace Ultraleap.TouchFree.Tooling.Connection
         // Also invokes <OnConnected> on all listeners.
         public void Connect()
         {
-            disconnected = false;
+            shouldReconnect = true;
 
             currentServiceConnection = new ServiceConnection(iPAddress, port, RetryConnecting);
             if (currentServiceConnection.IsConnected())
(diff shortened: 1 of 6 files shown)


TF_Application/Assets/TouchFree/Tooling/Scripts/Connection/ConnectionManager.cs
```diff
@@ -55,6 +55,8 @@ public static ServiceConnection serviceConnection
         // This value is settable in the Inspector.
         [SerializeField] string port = "9739";
 
+        static bool disconnected = true;
```
MattGrayUL: should this be 'connected' rather than 'disconnected'?

TF_Application/Assets/TouchFree/Tooling/Scripts/Connection/MessageReceiver.cs
```diff
@@ -105,6 +105,7 @@ void HandleResponse(WebSocketResponse _response)
                 if (callback.Key == _response.requestID)
                 {
                     callback.Value.callback.Invoke(_response);
+                    Debug.LogWarning(_response.message);
```
MattGrayUL, line 108: Does this need to be a warning? it could be a good response?

GabrielHomewoodUL: This change is already in the main tooling, I've just copied this to the Overlay application. I can swap this in both places if you think "Log" would be more appropriate?

TF_Application/Assets/TouchFree/Tooling/Scripts/Connection/ConnectionManager.cs
GabrielHomewoodUL: I'll rename this to something more appropriate

TF_Application/Assets/TouchFree/Tooling/Scripts/Connection/MessageReceiver.cs
MattGrayUL: I realised that after posting this! It's fine to leave for now then :)

Might be worth doing 2 commits in future to split the 'imported tooling' changes with the actual new elements :)

TF_Application/Assets/TouchFree/Tooling/Scripts/Connection/ConnectionManager.cs
MattGrayUL: I mean more to remove the idea of _wanting_ it to be false, I would typically prefer to know if I was connected and then fix it if I am _not_ connected, rather than looking for _not_ being disconnected

MattGrayUL (approved):

PR #270, opened 2022-04-21
(BUG) TF-488: Skip NONEs when catching up to a backlog of InputActions

NatGreenwood-UL:


TF_Application/Assets/TouchFree/Tooling/Scripts/Connection/MessageReceiver.cs
```diff
@@ -165,7 +165,8 @@ void CheckForAction()
                 if (actionQueue.TryPeek(out action))
                 {
                     // Stop shrinking the queue if we have a 'key' input event
-                    if (action.InputType != InputType.MOVE)
+                    if (action.InputType != InputType.MOVE ||
+                        action.InputType != InputType.NONE)
```
MattGrayUL: For all of these new if statements, I believe we need &&, not || otherwise it won't skip _anything_ ?

NatGreenwood-UL: Ah yeah we're doing NOT equal, my bad, I think soemthing similar worked the inverse way recently 🤦  Will fix shortly

MattGrayUL (approved):

PR #279, opened 2022-04-28
TF-498 - Tracking installer is left in the TouchFree directory after install

rblenkinsopp:
Sets the flag `deleteafterinstall` on the tracking installer so that the installation removes it after all installation steps have been completed.


MattGrayUL (approved):
I have tested the resulting build, the tracking installer does not remain after the install and uninstallation works as expected

Code makes sense, so happy to approve

PR #282, opened 2022-05-04
Add unit tests for utilities and small code move

GabrielHomewoodUL:
- [ ] Code Review 


TF_Service_dotNet/TouchFree/TouchFreeServiceTypes.cs
```diff
@@ -0,0 +1,128 @@
+using System;
+using Ultraleap.TouchFree.Library.Configuration;
+
+namespace Ultraleap.TouchFree.Library.ConnectionTypes
```
NatGreenwood-UL: If this is limited to Connection Types could we rename the file to match?
Also why are we putting _connection_ types (which are relevant only in the Service itself) into the core library? Is this to help with using the Library in Tooling?

GabrielHomewoodUL: This was a change that was meant to be backed out and had mostly been but this file was left behind. I have removed it now

NatGreenwood-UL (approved):

PR #294, opened 2022-06-15
Add an SVG cursor to the web settings UI

GabrielHomewoodUL:
- [x] Code Review 


TF_Settings_Web/public/index.html
```diff
@@ -27,6 +27,10 @@
     <!-- <script name="TouchFree" type="text/javascript" src="%PUBLIC_URL%/TouchFree/TouchFree_Tooling.js"></script> -->
   </head>
   <body>
+    <svg id="svg-cursor" class="touchfreecursor hidden" style="position: absolute; top: 0px; left: 0px; z-index: 1000" width="100%" height="100%" xmlns="http://www.w3.org/2000/svg">
```
PeterNancollis-UH: This doesn't need resolving now but does having this full screen svg region cause that whole area to repaint when the cursor moves? Just thinking there could be further performance gains if that's the case from making this area smaller and moving it around.

GabrielHomewoodUL: I'll add another task to investigate this and at the same time integrate the SVG Cursor into Web Tooling

PeterNancollis-UH (approved):

PR #296, opened 2022-06-17
Feature/tf 555 refactor interactions for flexibility

GabrielHomewoodUL:
- [x] Code Review 


TF_Service_dotNet/TouchFree/InteractionManager.cs
```diff
@@ -0,0 +1,177 @@
...
+                        if (interaction.Key == locationInteraction)
+                        {
+                            lastLocationActionToUpdate = interactionInputAction.inputAction;
+                            if (interactionCurrentlyDown == null)
+                            { 
+                                inputAction = interactionInputAction.inputAction;
+                            }
+                        }
+
+                        if (interactionCurrentlyDown == null && interactionInputAction != null && interactionInputAction.actionDetected && interactionInputAction.inputAction.InputType == InputType.DOWN)
```
NatGreenwood-UL, line 116: Doesn't the first part of this mean that we'll always favour whichever interactions happens to be earliest in the dict (if multiple are both experiencing downs)?

TF_Service_dotNet/TouchFree/Interactions/InteractionModules/AirPushInteraction.cs
```diff
@@ -110,7 +137,7 @@ private void HandleInteractionsAirPush()
                 Vector2 dPerp = virtualScreen.PixelsToMillimeters(dPerpPx);
 
                 // Update AppliedForce, which is the crux of the AirPush algorithm
-                float forceChange = GetAppliedForceChange(currentVelocity, dt, dPerp, distanceFromScreenMm);
+                float forceChange = GetAppliedForceChange(currentVelocity, dt, dPerp, distanceFromScreenMm) * confidence;  // Multiply by confidence to make it harder to use when disused
```
NatGreenwood-UL: Can we shift comments onto the line before rather than having super long lines? Minor but easier to read

TF_Service_dotNet/TouchFree/Interactions/InteractionModules/InteractionModule.cs
```diff
@@ -73,36 +95,19 @@ public void Update()
             {
                 handChirality = hand.IsLeft ? HandChirality.LEFT : HandChirality.RIGHT;
 
-                positions = positioningModule.CalculatePositions(hand);
+                positions = positioningModule.CalculatePositions(hand, positionConfiguration);
```
NatGreenwood-UL, line 98: This little trio of overwriting the same variable feels a little clunky (especially cos 2/3rds of the calls are in another object), could we pass a reference to the `ApplyAdditionalPositionModifiers` function into the positioning module call and have it do it all at once

TF_Service_dotNet/TouchFree/InteractionManager.cs
GabrielHomewoodUL: I don't think that this is a very big issue as it would require them both hitting down on the same frame. I was also thinking that we would add the interactions into the dictionary in priority order

TF_Service_dotNet/TouchFree/Interactions/InteractionModules/InteractionModule.cs
GabrielHomewoodUL: I could move that into the lower function, but there would be a similar setting of the variable in the positioningModule class. Ideally I think we would make the Positioning Module handle the PositionModifiers but then we would need a way to toggle the modifiers to use depending on the interaction which was a bit of additional complication that I think would be good to implement when we allow configuration of those items via settings

TF_Service_dotNet/TouchFree/Interactions/InteractionModules/AirPushInteraction.cs
GabrielHomewoodUL: Done!

TF_Service_dotNet/TouchFree/InteractionManager.cs
NatGreenwood-UL: I was wondering if we could trivially sort the list by the weight (if using weighted interactions) as well to lean in favour of this, but yes you're right because down happens on a _single frame_ this is very unlikely to occur. 👍

NatGreenwood-UL (approved):

PR #298, opened 2022-06-20
Feature/tf 557 add code analysis

GabrielHomewoodUL:
- [ ] Code Review 


NatGreenwood-UL:
2 comments on this one:

1. The "TouchFree" sonarqube lists a large number of duplicated lines, but these are all duplicated between TouchFree Web Tooling and the Web Settings UI. I wonder if we should exclude the TouchFree folder from within the Settings UI for now as it might obscure genuine duplications, and I think we already have work planned for addressing the "How do I use TFTfW as TypeScript" issue.

2. How do we maintain visibility of this info. I think the insights that we can get from SonarQube are great but they're not much use if they pop up exclusively outside of our regular flow. I think we would need something to actually provoke using this info and not just making there be more code smells and not noticing. I'd especially like if we can figure out some way to add the results to PR's the same way that the GitLab pipline has to pass for the "Check" below to pass

PR #301, opened 2022-06-24
TF-555 - Add swipes and filters

GabrielHomewoodUL:
Add swipes and filters

### DIFF
--- a/TF_Service_dotNet/TouchFree/Interactions/InteractionModules/VelocitySwipeInteraction.cs
+++ b/TF_Service_dotNet/TouchFree/Interactions/InteractionModules/VelocitySwipeInteraction.cs
@@ -267,17 +267,6 @@ namespace Ultraleap.TouchFree.Library.Interactions
             return false;
         }
 
-        protected override void OnInteractionSettingsUpdated(InteractionConfigInternal _config)
-        {
-            base.OnInteractionSettingsUpdated(_config);
-
-            //minScrollVelocity_mmps = _config.minScrollVelocity_mmps;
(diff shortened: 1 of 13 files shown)


TF_Service_dotNet/TouchFree/Interactions/InteractionModules/VelocitySwipeInteraction.cs
```diff
@@ -0,0 +1,296 @@
...
+            }
+
+            return false;
+        }
+
+        protected override void OnInteractionSettingsUpdated(InteractionConfigInternal _config)
+        {
+            base.OnInteractionSettingsUpdated(_config);
+
+            //minScrollVelocity_mmps = _config.minScrollVelocity_mmps;
```
NatGreenwood-UL: I _think_ these are all in the InteractionConfig file atm so do they need to be commented out?

NatGreenwood-UL:
PR looks good, going to try out the interactions locally and will throw any further feedback on here

TF_Service_dotNet/TouchFree/Interactions/InteractionModules/VelocitySwipeInteraction.cs
GabrielHomewoodUL: These were added before I created the interaction tuning json configuration file. I'll remove these as we will probably want a different way of configuring this interaction when we add the settings into the Settings UI

GabrielHomewoodUL: I have removed the references to these settings from the interaction configuration file now

NatGreenwood-UL (approved):

PR #302, opened 2022-06-27
Feature/tf 561 add quick setup functionality to service

GabrielHomewoodUL:
- [ ] Code Review 


TF_Service_dotNet/TouchFree/Configuration/QuickSetupHandler.cs
```diff
@@ -0,0 +1,123 @@
...
+
+        public QuickSetupHandler(IHandManager _handManager, ITrackingConnectionManager _trackingConnectionManager, IConfigManager _configManager)
+        {
+            handManager = _handManager;
+            trackingConnectionManager = _trackingConnectionManager;
+            configManager = _configManager;
+        }
+
+        public void HandleQuickSetupCall(QuickSetupPosition position)
+        {
```
NatGreenwood-UL: It doesn't seem like there's a system for responding to the Client here; generally our requests respond to the client with a short response to indicate success (or otherwise). I think particularly here this system should return a message that indicates that they've done something wrong if recieving a bottom position before a top, as if that will be ignored by the service, clients _need_ a way to know that information.

NatGreenwood-UL (commented):
I think we need some sort of response from the service here to indicate success/failure, particularly because if we don't there's a silent failure state.

Also I realise it was listed as out of scope on the ticket, but I was expecting this work to include updating Unity **tooling**, just not the Unity Settings UI. Any idea why we shouldn't do that now? If we don't, Unity Tooling ends up in a state where its API is out of sync with the Service.

TF_Service_dotNet/TouchFree_Service/Connection/Services/WebSocketReceiver.cs
```diff
@@ -132,17 +132,30 @@ void HandleQuickSetupCall(string _content)
                 return;
             }
 
-            quickSetupHandler.HandleQuickSetupCall(quickSetupRequest.Value.Position);
+            var quickSetupResponse = quickSetupHandler.HandleQuickSetupCall(quickSetupRequest.Value.Position);
```
NatGreenwood-UL: This is a call to an indentically named function in another class, could the one here be renamed to fit more with the other functions in the class (e.g. `HandleQuickSetupRequest`)

TF_Service_dotNet/TouchFree_Service/Connection/Services/WebSocketReceiver.cs
```diff
@@ -132,17 +132,30 @@ void HandleQuickSetupCall(string _content)
...
+                clientMgr.SendQuickSetupConfigFile(currentConfig);
+            }
+            else if (quickSetupResponse?.PositionRecorded == true)
+            {
+                ResponseToClient response = new ResponseToClient("", "Success", "", _content);
+                clientMgr.SendQuickSetupResponse(response);
+            }
+            else
+            {
+                ResponseToClient response = new ResponseToClient("", "Failure", "Unable to record hand position", _content);
```
NatGreenwood-UL: Could we make this error message clear with the *cause* of this problem. 

The scenario I'm considering is someone who's not really read the docs trying to use Tooling to create their own quick setup, not knowing that they have to do Top First, and then getting an error that doesn't give them a way to move forward.
I believe that the only ways this can fail are if there's no tracked hand or if they're being requested in the wrong order; I think distinguishing between those errors is going to make the feature far more usable in future.

NatGreenwood-UL (approved):
Looks good! Might be good to get Jordan to re-pull this into the UI branch to test the flow before merging _just_ for safety but otherwise 👌

PR #305, opened 2022-07-04
Bug/tf 570 primary hand updating

GabrielHomewoodUL:
- [ ] Code Review 


TF_Service_dotNet/TouchFree/HandManager.cs
```diff
@@ -182,12 +197,86 @@ public void Update(object sender, FrameEventArgs e)
...
         }
 
+        private bool HandActivityInSwapThreshold()
+        {
+            return PrimaryHandActivity < SecondaryHandActivityDefault && SecondaryHandActivity > PrimaryHandActivityDefault;
+        }
+
+        private static bool PrimaryHandIsOffScreen(float _primaryRelativeXScreenPosition, float _primaryRelativeYScreenPosition)
+        {
+            return _primaryRelativeXScreenPosition > 1.4 ||
```
NatGreenwood-UL, line 282: If I'm reading this right does this mean that for a hand to be considered "off-screen" it needs to be more than 40% of the screen's size in that dimension outside the screen? So "on-screen" is an area about 180% of the width and 180% of height vs. the _actual_ screen size?

NatGreenwood-UL (approved):

TF_Service_dotNet/TouchFree/HandManager.cs
GabrielHomewoodUL: Yeah, that's right, we could potentially reduce that but I don't think it would be good to have it the same as the screen size as it could make clicking buttons near the edge of the screen a pain

NatGreenwood-UL: I agree on not making it a 1-1 relationship, I'm just wondering about the 40% either side? Might be good to get some feedback from Chris on how far this might go. I think that with 40% (especially with larger screens) you're still quite likely to have the "other hand holding coffee" issue quite easily?

PR #309, opened 2022-07-11
TF-545 - Update to switch air push to have a narrower code when scroll and drag is off

GabrielHomewoodUL:
- [x] Code Review 

### DIFF
--- a/TF_Service_dotNet/TouchFree/Interactions/InteractionModules/AirPushInteraction.cs
+++ b/TF_Service_dotNet/TouchFree/Interactions/InteractionModules/AirPushInteraction.cs
@@ -1,9 +1,7 @@
-﻿using System;
-using System.Collections.Generic;
+﻿using Microsoft.Extensions.Options;
+using System;
 using System.Diagnostics;
 using System.Numerics;
-using Microsoft.Extensions.Options;
-using Ultraleap.TouchFree.Library;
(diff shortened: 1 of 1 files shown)


TF_Service_dotNet/TouchFree/Interactions/InteractionModules/AirPushInteraction.cs
```diff
@@ -91,6 +89,13 @@ public class AirPushInteraction : InteractionModule
             };
         }
 
+        protected override void OnInteractionSettingsUpdated(InteractionConfigInternal _config)
+        {
+            base.OnInteractionSettingsUpdated(_config);
+
+            thetaOne = ignoreDragging ? 15f : 65f;
```
PeterNancollis-UH, line 96: The narrower angle is set to 15 degrees, but looking at the previous default value above it was 22 degrees. Is this additional narrowing on purpose?

GabrielHomewoodUL: I don't think it was ever meant to be 22 degrees. That was the angle that Chris used when testing with swipes, before we made the changes for scroll and drag it was 15 degrees

PeterNancollis-UH (approved):

PR #320, opened 2022-07-19
Merge web tooling changes into main feature branch for TF-571

cartmouse:
- [ ] Code Review
- [ ] Developer Testing


TF_Tooling_Web/src/Connection/ServiceConnection.ts
```diff
@@ -268,4 +276,24 @@ export class ServiceConnection {
 
         this.webSocket.send(message);
     }
+
+    // Function: RequestTrackingState
+    // Requests 
+    RequestTrackingState(_callback: (detail: TrackingStateResponse) => void) {
+        if (!_callback) {
+            console.error('Request failed. This is due to a missing callback');
```
NatGreenwood-UL: We've got quite a few of these now, wonder if we should be improving this message to clarify which call causes the issue (especially as the function doesn't _actually_ error)

NatGreenwood-UL (commented):
Just one minor comment, could you also merge the update to the base branch into this so I can test with the Tooling Test App repo?

TF_Tooling_Web/src/Connection/ServiceConnection.ts
cartmouse: A good shout, I'll make these error messages a bit more specific to the calls they're making.

NatGreenwood-UL (approved):
I've now used this in the TF Test Application to check that the flow through both this, service and onto Tracking system, which all worked great! Happy for this to be merged into the main 571 branch now, I'll get the Unity Tooling done there also

PR #331, opened 2022-07-28
TF-616: Add Hand Renders to WebSettings UI

GabrielHomewoodUL:
- [ ] Code Review
- [x] Developer Testing
- [ ] QA Review
- [ ] Test Plan updated
- [ ] PO Review (Optional)
- [ ] Update Release Notes


TF_Settings_Web/src/Components/Controls/HandsSvg.tsx
```diff
@@ -0,0 +1,213 @@
...
+
+    wrist: HandSvgCoordinate;
+
+    primaryHand: boolean;
+
+    dotColor: string;
+}
+
+interface DataWrapper {
+    one: HandSvgProps | undefined;
```
JordanBerni-UL: Adding a ? makes the member optional instead of needing to use | undefined (e.g. one? : HandSvgProps)

TF_Settings_Web/src/Components/Controls/HandsSvg.tsx
```diff
@@ -0,0 +1,213 @@
...
+            </defs>
+
+            <HandSvg data={one} />
+            <HandSvg data={two} />
+        </svg>
+    );
+};
+
+interface HandDataWrapper {
+    data: HandSvgProps | undefined;
```
JordanBerni-UL: data?: HandSvgProps

TF_Settings_Web/src/Components/Pages/Camera/Masking/MaskingScreen.tsx
```diff
@@ -2,32 +2,45 @@ import 'Styles/Camera/CameraMasking.scss';
...
 import SwapMainLensIcon from 'Images/Camera/Swap_Main_Lens_Icon.svg';
 
+import { HandsSvg, HandSvgProps } from 'Components/Controls/HandsSvg';
+
 import MaskingOption from './MaskingOptions';
 import MaskingSlider, { SliderDirection } from './MaskingSlider';
 import { displayLensFeeds } from './displayLensFeeds';
+import { defaultHandState, handToSvgData, setHandRenderState } from './handRendering';
+
+interface HandRenderState {
```
JordanBerni-UL: Same as DataWrapper - could these be the same interface (with a different name) and export it across the two files?

TF_Settings_Web/src/Components/Pages/Camera/Masking/MaskingScreen.tsx
```diff
@@ -2,32 +2,45 @@ import 'Styles/Camera/CameraMasking.scss';
...
+    handTwo: HandSvgProps | undefined;
+}
 
 enum Lens {
     Left,
     Right,
 }
 
 const MaskingScreen = () => {
+    const [handData, setHandData] = useState<HandRenderState>(defaultHandState);
```
JordanBerni-UL: Should this not come under the state heading?

TF_Settings_Web/src/Components/Pages/Camera/Masking/MaskingScreen.tsx
```diff
@@ -78,38 +103,65 @@ const MaskingScreen = () => {
...
             return;
         }
 
-        setIsFrameProcessing(true);
+        const dataAsUint8 = new Uint8Array(event.data, 0, 10);
 
-        if (!successfullySubscribed.current) {
+        if (dataAsUint8[0] === 1) {
+            successfullySubscribed.current = true;
+            setTimeout(() => {
```
JordanBerni-UL: Why so many nested timeouts? To make it async? Is it worth looking into a better way to do this? promises / async + await?

TF_Settings_Web/src/Components/Pages/Camera/Masking/MaskingScreen.tsx
```diff
@@ -78,38 +103,65 @@ const MaskingScreen = () => {
...
+                if (handOne) {
+                    convertedHandOne = handToSvgData(handOne, 0);
+                }
+                if (handTwo) {
+                    convertedHandTwo = handToSvgData(handTwo, 1);
+                }
+                setHandData({ handOne: convertedHandOne, handTwo: convertedHandTwo });
+            }
+
+            // Settimeout with 32ms for ~30fps if we have the performance
```
JordanBerni-UL: out dated comment

TF_Settings_Web/src/Components/Pages/Camera/Masking/MaskingScreen.tsx
```diff
@@ -78,38 +103,65 @@ const MaskingScreen = () => {
...
+            setTimeout(() => {
+                isHandProcessingRef.current = false;
+            }, 67);
```
JordanBerni-UL: variable here to keep it in sync with the camera FPS?

TF_Settings_Web/src/Components/Pages/Camera/Masking/MaskingScreen.tsx
```diff
@@ -78,38 +103,65 @@ const MaskingScreen = () => {
...
 
-        // Settimeout with 32ms for ~30fps if we have the performance
+        isHandProcessingRef.current = true;
         setTimeout(() => {
-            setIsFrameProcessing(false);
-        }, 32);
+            if (evt.detail?.Hands) {
+                const handOne = evt.detail.Hands[0];
+                const handTwo = evt.detail.Hands[1];
+                let convertedHandOne: HandSvgProps | undefined = undefined;
```
JordanBerni-UL: Optional: could do
`const convertedHandOne = handOne ? handToSvgData(handOne, 0) : undefined;`

(think this looks a lot cleaner)

TF_Settings_Web/src/Components/Pages/Camera/Masking/displayLensFeeds.ts
```diff
@@ -27,44 +29,97 @@ export const displayLensFeeds = (
...
-    rightLensRef.height = lensHeight;
-    rightContext.putImageData(new ImageData(rightBuf8, width, lensHeight), 0, 0);
+export const getContext = (canvasElement: HTMLCanvasElement | null) => {
+    if (!canvasElement) {
+        return undefined;
+    }
+    const canvasIndex = canvasElements.indexOf(canvasElement);
+    if (canvasIndex < 0) {
+        const context = canvasElement.getContext('2d');
+        if (context == null) {
```
JordanBerni-UL: Only double equals is a ts sin!

TF_Settings_Web/src/Components/Pages/Camera/Masking/displayLensFeeds.ts
```diff
@@ -27,44 +29,97 @@ export const displayLensFeeds = (
...
+            return context;
```
JordanBerni-UL: If !canvasElement we return undefined but if !context we return null but don't seem to do anything with this differentiated information so could we return the same in both to limit the number of return types?

TF_Settings_Web/src/Components/Pages/Camera/Masking/displayLensFeeds.ts
```diff
@@ -27,44 +29,97 @@ export const displayLensFeeds = (
...
+        }
+        const newIndex = canvasElements.length;
```
JordanBerni-UL, line 116: React has caching inbuilt using memoization (React.memo) to avoid having to implement your own (would need to check that this works with canvasElement as it may be too complex of an input? (but probably not))

TF_Settings_Web/src/Components/Pages/Camera/Masking/handRendering.ts
```diff
@@ -0,0 +1,90 @@
+import { v4 as uuidgen } from 'uuid';
```
JordanBerni-UL, line 1: Side note but might want to add this to the prettier import order regex so we can control where to put it (not really important though)

TF_Settings_Web/src/Components/Pages/Camera/Masking/handRendering.ts
```diff
@@ -0,0 +1,90 @@
...
+    const request = new CommunicationWrapper(ActionCode.SET_HAND_DATA_STREAM_STATE, content);
+
+    const jsonContent = JSON.stringify(request);
+
+    ConnectionManager.serviceConnection()?.SendMessage(jsonContent, requestID, () => {
+        return true;
+    });
+};
+
+const translateToCoordinate = (coordinate: Vector | undefined) => {
```
JordanBerni-UL: `coordinate?: Vector`

TF_Settings_Web/src/Components/Pages/Camera/Masking/handRendering.ts
```diff
@@ -0,0 +1,90 @@
...
+        middleKnuckle: translateToCoordinate(middleFinger?.Bones[knuckleJointIndex]?.PrevJoint),
+        ringTip: translateToCoordinate(ringFinger?.Bones[tipJointIndex]?.NextJoint),
+        ringKnuckle: translateToCoordinate(ringFinger?.Bones[knuckleJointIndex]?.PrevJoint),
+        littleTip: translateToCoordinate(littleFinger?.Bones[tipJointIndex]?.NextJoint),
+        littleKnuckle: translateToCoordinate(littleFinger?.Bones[knuckleJointIndex]?.PrevJoint),
+        thumbTip: translateToCoordinate(thumbFinger?.Bones[tipJointIndex]?.NextJoint),
+        thumbKnuckle: translateToCoordinate(thumbFinger?.Bones[knuckleJointIndex]?.PrevJoint),
+        wrist: translateToCoordinate(wrist),
+        primaryHand: hand.CurrentPrimary,
+        dotColor: handIndex ? 'blue' : 'red',
```
JordanBerni-UL: Hand index being a number but being used as a bool -> could make this an enum or a type?

TF_Settings_Web/src/Styles/Camera/CameraMasking.scss
```diff
@@ -15,11 +15,26 @@
...
 
-    > canvas {
+    > div {
+        $canvas-size: $size - $padding;
+        margin-top: calc((100% - $canvas-size) / 2);
+        display: inline-block;
+        z-index: 10;
+    }
+
+    > div.cam-feed-box-hand-renders {
```
JordanBerni-UL: these could all go in the > div {}  (optional - depends on preference) 

> 
> '>' div {
>   &.cam-feed-box-hand-renders{}
>   '>' canvas {}
> }

JordanBerni-UL: '>' is just so git doesn't format the >

TF_Settings_Web/src/TouchFree/Connection/MessageReceiver.ts
```diff
@@ -223,6 +229,30 @@ export class MessageReceiver {
             this.lastStateUpdate = HandPresenceState.PROCESSED;
         }
     }
+    
+    // Function: CheckForAction
```
JordanBerni-UL: outdated comment

TF_Settings_Web/src/TouchFree/Connection/TouchFreeServiceTypes.ts
```diff
@@ -136,6 +140,21 @@ export class ConfigState extends TouchFreeRequest {
 export class ConfigChangeRequest extends TouchFreeRequest {
 }
 
+// class: ConfigChangeRequest
```
JordanBerni-UL: outdated comment

TF_Settings_Web/src/TouchFree/Plugins/HandDataManager.ts
```diff
@@ -0,0 +1,55 @@
+import { HandFrame } from "TouchFree/TouchFreeToolingTypes";
+import { InputActionPlugin } from "./InputActionPlugin";
+
+// Class: InputActionManager
```
JordanBerni-UL: comment

TF_Settings_Web/src/TouchFree/Plugins/HandDataManager.ts
```diff
@@ -0,0 +1,55 @@
...
+// The manager for all <TouchFreeInputActions> to be handled and distributed. This runs the
+// received data through any <InputActionPlugins> given to it and finaly distributes the data
+// via the  <TransmitInputAction> event which should be listened to by any class hoping to make
+// use of incoming <TouchFreeInputActions>.
+export class HandDataManager extends EventTarget {
+    // Event: TransmitHandData
+    // An event for transmitting <TouchFreeInputActions> that are received via the <messageReceiver> to
+    // be listened to.
+
+    // Event: TransmitInputActionRaw
```
JordanBerni-UL: comment

TF_Settings_Web/src/TouchFree/Plugins/HandDataManager.ts
```diff
@@ -0,0 +1,55 @@
+import { HandFrame } from "TouchFree/TouchFreeToolingTypes";
+import { InputActionPlugin } from "./InputActionPlugin";
+
+// Class: InputActionManager
+// The manager for all <TouchFreeInputActions> to be handled and distributed. This runs the
+// received data through any <InputActionPlugins> given to it and finaly distributes the data
+// via the  <TransmitInputAction> event which should be listened to by any class hoping to make
+// use of incoming <TouchFreeInputActions>.
+export class HandDataManager extends EventTarget {
+    // Event: TransmitHandData
```
JordanBerni-UL, line 10: comment

JordanBerni-UL (changes requested):

TF_Tooling_Web/src/Connection/MessageReceiver.ts
```diff
@@ -7,10 +7,9 @@ import {
     ServiceStatus,
     ServiceStatusCallback,
     TouchFreeRequest,
-    TrackingStateRequest,
     TrackingStateCallback,
     TrackingStateResponse,
-    WebSocketResponse
+    WebSocketResponse,
```
JordanBerni-UL, line 12: This change seems out of place in this PR as it's the only TF_Tooling_Web change

TF_Settings_Web/src/Components/Pages/Camera/Masking/displayLensFeeds.ts
```diff
@@ -27,44 +29,97 @@ export const displayLensFeeds = (
...
+    lensHeight: number
+) => {
+    lensRef.width = width;
+    lensRef.height = lensHeight;
+    context.putImageData(new ImageData(buf8, width, lensHeight), 0, 0);
+};
 
-        for (let i = 0; i < width * lensHeight; i++) {
-            rightBuf32[i] = byteConversionArray[offsetView.getUint8(i)];
+export const processRotatedScreen = (
```
JordanBerni-UL: This is only used within this file - does it need export? (I remember you saying about using this for profiling reasons but can this be removed now)

TF_Settings_Web/src/TouchFree/Plugins/HandDataManager.ts
```diff
@@ -0,0 +1,41 @@
...
+import { InputActionPlugin } from "./InputActionPlugin";
+
+// Class: HandDataManager
+// The manager for all <HandFrame> to be handled and distributed. This distributes the data
+// via the <TransmitHandData> event which should be listened to by any class hoping to make
+// use of incoming <HandFrame>s.
+export class HandDataManager extends EventTarget {
+    // Event: TransmitHandData
+    // An event for transmitting <HandFrame>s that are received via the <messageReceiver> to
+    // be listened to.
```
JordanBerni-UL, line 10: Should this comment be here? Confused as to what Event it is describing

TF_Settings_Web/src/TouchFree/Plugins/HandDataManager.ts
```diff
@@ -0,0 +1,41 @@
+import { HandFrame } from "TouchFree/TouchFreeToolingTypes";
+import { InputActionPlugin } from "./InputActionPlugin";
```
JordanBerni-UL: Unused import

TF_Settings_Web/src/TouchFree/Plugins/HandDataManager.ts
GabrielHomewoodUL: This event is triggered lower down, it's how we document the events as they don't really exist like a property within JS

TF_Tooling_Web/src/Connection/MessageReceiver.ts
GabrielHomewoodUL: It's just a bit of clean up so I was thinking of just leaving this

JordanBerni-UL: That's fine - just wanted to check it was intentional

JordanBerni-UL (approved):

TF_Settings_Web/src/Components/Controls/HandsSvg.tsx
```diff
@@ -0,0 +1,231 @@
...
+            </defs>
+
+            <HandSvg data={one} />
+            <HandSvg data={two} />
+        </svg>
+    );
+};
+
+export const HandSvg: React.FC<{ data?: HandSvgProps }> = ({ data }) => {
+    if (!data?.dotColor) {
```
cartmouse: Why are we checking for dotColor specifically here? Would it not be better to do `if (!data) ...`?

JordanBerni-UL: Yeah agreed - the HandSVG doesn't seem to use dot color at the moment (Gabe might have put it in for future use) so maybe this was to stop the unused variable warning?

JordanBerni-UL: Ah as it isn't unpacked from the data variable it doesn't give unused variable warning so is all good

JordanBerni-UL: Changed 👍

cartmouse: Nice, looks good!

TF_Service_dotNet/TouchFree/Connections/MessageQueues/QuickSetupQueueHandler.cs
```diff
@@ -0,0 +1,66 @@
...
+        {
+            QuickSetupRequest? quickSetupRequest = null;
+
+            try
+            {
+                quickSetupRequest = JsonConvert.DeserializeObject<QuickSetupRequest>(_request.content);
+            }
+            catch { }
+
+            // Explicitly check for requestID because it is the only required key
```
NatGreenwood-UL: As most all inheritors from MessageQueueHandler have to do this requestID check, could we move it inside the base? I think the only variation between them is the message used in the response, so could we override a String on the child classes that could then be used in the base?

NatGreenwood-UL: I suppose the method for handling this would need to be overwritable though, as the TrackingAPIChangeHandler needs to respond with an entirely different type

NatGreenwood-UL (commented):
Just one comment on the refactor about duplicated method, otherwise the Svc side of this looks great 👌

TF_Service_dotNet/TouchFree/Connections/ActionCode.cs
```diff
@@ -0,0 +1,83 @@
+﻿using System.Linq;
+
+namespace Ultraleap.TouchFree.Library.Connections
+{
+    public enum ActionCode
+    {
+        INPUT_ACTION,
```
Craig-J, line 7: Give enum values explicit numbers so that inserting new members anywhere except the end is not breaking

TF_Service_dotNet/TouchFree/Connections/ClientConnection.cs
```diff
@@ -178,52 +194,22 @@ internal void OnMessage(string _message)
...
-
-            switch (action)
+            var queueHandler = messageQueueHandlers.SingleOrDefault(x => x.ActionCodes.Contains(action));
+            if (queueHandler != null)
+            {
+                queueHandler.AddItemToQueue(new IncomingRequest(action, null, content));
+            }
+            else if (action.ExpectedToBeHandled())
+            {
+                TouchFreeLog.ErrorWriteLine("Expected to be able to handle a " + action + " action but unable to find queue.");
```
Craig-J: Use string interpolation (won't mention this more than once, else there'd be loads of comments)

TF_Service_dotNet/TouchFree/Connections/ClientConnectionManager.cs
```diff
@@ -0,0 +1,223 @@
...
+
+namespace Ultraleap.TouchFree.Library.Connections
+{
+    public class ClientConnectionManager : IClientConnectionManager
+    {
+        private readonly ConcurrentDictionary<Guid, IClientConnection> activeConnections = new ConcurrentDictionary<Guid, IClientConnection>();
+
+        public IEnumerable<IClientConnection> ClientConnections
+        {
+            get { return activeConnections.Values; }
```
Craig-J: Since this is a small simple single expression function, consider using an expression body `public IEnumerable<IClientConnection> ClientConnections => activeConnections.Values;`
Won't mention in future as it's a formatting issue and may create a lot of review noise

TF_Service_dotNet/TouchFree/Connections/ClientConnectionManager.cs
```diff
@@ -0,0 +1,223 @@
...
+        }
+
+        public event Action LostAllConnections;
+
+        public short port = 9739;
+
+        public HandPresenceEvent MissedHandPresenceEvent { get; private set; }
```
Craig-J, line 20: Use `event` keyword. Currently this event's invocation list is mutable from the public API and it's also possible to invoke from outside of an instance - either of these behaviours from API users can lead to difficult to debug bugs. `event` fixes both of these by changing the public API to only add/remove subscription functions, only allowing invocation and direct modification of the subscribers privately. There are also thread synchronization concerns when invoking a non `event` delegate. See [here](https://stackoverflow.com/questions/18385967/c-sharp-event-keyword-advantages) for more details.

TF_Service_dotNet/TouchFree/Connections/ClientConnectionManager.cs
```diff
@@ -0,0 +1,223 @@
...
+            {
+                MissedHandPresenceEvent = handsLostEvent;
+            }
+        }
+
+        public void AddConnection(IClientConnection _connection)
+        {
+            if (_connection != null)
+            {
+                activeConnections.TryAdd(Guid.NewGuid(), _connection);
```
Craig-J, line 65: Unused return? What happens if this fails?

TF_Service_dotNet/TouchFree/Connections/ClientConnectionManager.cs
```diff
@@ -0,0 +1,223 @@
...
+            });
+        }
+
+        public void SendHandDataStreamStateResponse(ResponseToClient _response)
+        {
+            SendMessageToWebSockets((connection) =>
+            {
+                connection.SendHandDataStreamStateResponse(_response);
+            });
+        }
```
Craig-J: These are all just wrappers to connection calls of the same name... this might indicate some kind of architectural issue?

TF_Service_dotNet/TouchFree/Connections/ConnectionMessageTypes.cs
```diff
@@ -73,6 +32,54 @@ public ConfigState(string _id, InteractionConfig _interaction, PhysicalConfig _p
...
+        public TrackingServiceState trackingServiceState;
+        public ConfigurationState configurationState;
+
+        public ServiceStatus(string _id, TrackingServiceState _trackingServiceState, ConfigurationState _configurationState)
+        {
+            requestID = _id;
+            trackingServiceState = _trackingServiceState;
+            configurationState = _configurationState;
+        }
+    }
```
Craig-J, line 48: Use record type here as it's less boilerplate.
`public record ServiceStatus(string requestID, TrackingServiceState trackingServiceState, ConfigurationState configurationState);`
Will not mention this suggestion later in review for other cases

TF_Service_dotNet/TouchFree/Connections/ConnectionTypes.cs
```diff
@@ -0,0 +1,23 @@
+﻿namespace Ultraleap.TouchFree.Library.Connections
+{
+    public enum Compatibility
+    {
+        COMPATIBLE,
```
Craig-J, line 5: Add explicit numbering to enum members for same reason as before

TF_Service_dotNet/TouchFree/Connections/ConnectionTypes.cs
```diff
@@ -0,0 +1,23 @@
...
+        SERVICE_OUTDATED,
+        CLIENT_OUTDATED,
+        SERVICE_OUTDATED_WARNING,
+        CLIENT_OUTDATED_WARNING
+    }
+
+    internal struct CommunicationWrapper<T>
```
Craig-J, line 12: Use generic record `internal record CommunicationWrapper<T>(string action, T content);`

TF_Service_dotNet/TouchFree/Connections/IClientConnection.cs
```diff
@@ -1,9 +1,24 @@
...
+        void SendConfigFileChangeResponse(ResponseToClient response);
+        void SendConfigState(ConfigState config);
+        void SendConfigFile(ConfigState config);
+        void SendStatusResponse(ResponseToClient response);
+        void SendStatus(ServiceStatus response);
+        void SendQuickSetupConfigFile(ConfigState config);
+        void SendQuickSetupResponse(ResponseToClient response);
+        void SendTrackingState(TrackingApiState state);
+        void SendHandDataStreamStateResponse(ResponseToClient response);
+        void SendHandPresenceEvent(HandPresenceEvent handsLostEvent);
```
Craig-J: Seems to mostly duplicate what was added to `IClientConnectionManager` - should there be a shared interface here? The duplication of signatures makes me think there's likely a better way to solve this architecturally since these interfaces are across different levels of the API. Should the manager return specific connections? e.g. `IClientConnection GetConnection(Guid id)` instead of duplicating the API? Would like to discuss this in more details.

TF_Service_dotNet/TouchFree/Connections/MessageQueues/BaseConfigurationChangeQueueHandler.cs
```diff
@@ -0,0 +1,150 @@
...
+        /// <summary>
+        /// Checks for invalid states of the config request
+        /// </summary>
+        /// <param name="_content">The whole json content of the request</param>
+        /// <returns>Returns a response as to the validity of the _content</returns>
+        protected ResponseToClient ValidateConfigChange(string _content)
+        {
+            ResponseToClient response = new ResponseToClient(string.Empty, "Success", string.Empty, _content);
+
+            JObject contentObj = JsonConvert.DeserializeObject<JObject>(_content);
```
Craig-J: Create a data type to represent the structure of the deserialized object instead of using reflection to check specific fields. If this doesn't fit our needs I think it'd be preferrable to explore other Json.NET APIs to control deserialization instead of inspecting fields with reflection

TF_Service_dotNet/TouchFree/Connections/IClientConnection.cs
GabrielHomewoodUL: This is something I would like to change but didn't want this PR to get any bigger! I think that we should just have a generic send method in clientconnection and have the width of methods in client connection manager but one for a tidy up at another time.

TF_Service_dotNet/TouchFree/Connections/MessageQueues/BaseConfigurationChangeQueueHandler.cs
GabrielHomewoodUL: I think this would be a beneficial change but I didn't want to change too much in this PR and this code was directly moved from the MessageReceiver class

TF_Service_dotNet/TouchFree/Connections/ClientConnectionManager.cs
GabrielHomewoodUL: This isn't a c# event. It's a way of recording the hands lost/found event that we propagate to clients incase of an unexpected disconnection so that on reconnection we let them know the current hand presence state. It could maybe be renamed to avoid confusion

TF_Service_dotNet/TouchFree/Connections/MessageQueues/QuickSetupQueueHandler.cs
GabrielHomewoodUL: I've had a decent look at this and I can't really come up with a good way to do this without adding a lot more code. I'd be interested to discuss for suggestions

TF_Service_dotNet/TouchFree/Connections/ClientConnectionManager.cs
GabrielHomewoodUL: I think this is valid to mention. I prefer the syntax you suggested to use, I think I just missed swapping to use it during progressive edits to the code

TF_Service_dotNet/TouchFree/Connections/ConnectionMessageTypes.cs
GabrielHomewoodUL: I think it would be good to do this as a bit of tech debt as most of these have just been moved and I'm reluctant to make more broad changes

TF_Service_dotNet/TouchFree/Connections/ClientConnectionManager.cs
GabrielHomewoodUL: I think it's ok to ignore this if it fails, we will retry removing it on the next message that we send

TF_Service_dotNet/TouchFree/Connections/ClientConnectionManager.cs
GabrielHomewoodUL: Yes, I would like to change this but I'm reluctant to make more broad changes within this PR

TF_Service_dotNet/TouchFree/Connections/ActionCode.cs
GabrielHomewoodUL: These are only ever exposed outside of the application as strings so we shouldn't have this issue. They should probably be moved away from being public in this case though

TF_Service_dotNet/TouchFree/Connections/ClientConnection.cs
GabrielHomewoodUL: I am happy to move string manipulation to use interpolation, I don't want to spend too much time rewriting stuff though. It might be worth us putting together some linting and doing a big tech debt piece to cover items like this

TF_Service_dotNet/TouchFree/Connections/MessageQueues/BaseConfigurationChangeQueueHandler.cs
NatGreenwood-UL: There is actually an underlying type for these changes (the `PhysicalConfig` and `InteractionConfig` types) but we do this piecemeal manual deserialisation deliberately to avoid an issue with how Newtonsoft parses partial types.

These types are partial by design so that you can essentially pass a "patch" of changes you want made and not include any other members of the config types. Unfortunately, if you attempt to parse a partial change through Newtonsoft, it _works_ but it will fill in all the missing values with their defaults. As we don't want to change anything which wasn't modified deliberately by the sender, and we can't trust that they weren't trying to send the default value, this parsing doesn't meet our needs.
Unfortunately we also cannot get around this by using `Nullable` entries for everything on the type as Newtonsoft will parse anything Nullable as nonexistant (even if it is present in the JSON string).

TF_Service_dotNet/TouchFree/Connections/ClientConnectionManager.cs
Craig-J: Ah ok - should it be private then? Seems like it shouldn't be something exposed by the manager if it's for bookkeeping

GabrielHomewoodUL: It's consumed by the ClientConnection class so needs to be accessible over the IClientConnectionManager interface so can't be private

NatGreenwood-UL (approved):

TF_Service_dotNet/TouchFree/Connections/ActionCode.cs
GabrielHomewoodUL: Added a task for this to the techdebt epic. Probably one to discuss within the team

TF_Service_dotNet/TouchFree/Connections/ClientConnection.cs
GabrielHomewoodUL: Added a techdebt task for this to cover other areas where we want to do this

TF_Service_dotNet/TouchFree/Connections/ClientConnectionManager.cs
GabrielHomewoodUL: I think the interaction between client connection and client connection manager could be improved and simplified but that would be a story in itself

TF_Service_dotNet/TouchFree/Connections/ConnectionMessageTypes.cs
GabrielHomewoodUL: Added a techdebt task for this

TF_Service_dotNet/TouchFree/Connections/ConnectionTypes.cs
GabrielHomewoodUL: Covered above

TF_Service_dotNet/TouchFree/Connections/ConnectionTypes.cs
GabrielHomewoodUL: Covered above

TF_Service_dotNet/TouchFree/Connections/IClientConnection.cs
GabrielHomewoodUL: I've simplified this a lot and I've also added another task to work on this further

TF_Service_dotNet/TouchFree/Connections/MessageQueues/BaseConfigurationChangeQueueHandler.cs
GabrielHomewoodUL: I've added a task to see if we can do any further improvements to this (though it's more of a spike)

TF_Service_dotNet/TouchFree/Connections/ClientConnectionManager.cs
GabrielHomewoodUL: I ended up resolving the large majority of these but some further work could be done

PR #334, opened 2022-08-02
TF-572: Update config via Camera Masking Screen

JordanBerni-UL:
- [x] Code Review
- [x] Developer Testing
- [x] QA Review
- [x] Test Plan updated
- [x] PO Review (Optional)
- [x] Update Release Notes

**Known Issues**

- Sliders only update on drag end
- Camera's sometimes flicker

Created https://ultrahaptics.atlassian.net/browse/TF-649 to look at BRS performance issues


TF_Settings_Web/src/Components/Controls/ToggleSwitch.tsx
```diff
@@ -4,32 +4,26 @@ import React from 'react';
...
-        <input
-            type="checkbox"
-            style={{ display: 'none' }}
-            checked={value}
-            onChange={() => {
-                onChange(!value);
-            }}
-        />
+        {/* eslint-disable-next-line @typescript-eslint/no-empty-function */}
+        <input type="checkbox" style={{ display: 'none' }} checked={value} onChange={() => {}} />
```
NatGreenwood-UL, line 14: Can these ToggleSwitches no longer be changed? Seems like they wouldn't do anything on click now?

TF_Settings_Web/src/TouchFree/Configuration/ConfigurationManager.ts
```diff
@@ -62,14 +62,14 @@ export class ConfigurationManager {
     public static RequestConfigFileChange(
         _interaction: Partial<InteractionConfig> | null,
```
NatGreenwood-UL, line 63: All the changes to the TouchFree Tooling applied from here down in the listed changes need to be replicated across to the "main" version of the TouchFree tooling before we end up with really divergent versions.

NatGreenwood-UL: Copying one atop the other they seem to have quite a few differences: 
image

TF_Settings_Web/src/Components/Controls/ToggleSwitch.tsx
JordanBerni-UL: The label around the ToggleSwitch takes care of what to do when the checkbox is clicked (TF doesn't trigger the onChange function and that would only apply to the actual switch element, not the surrounding label).

I will add this as a comment to make it more clear

cartmouse:
I've had a look over these latest changes and it all looks good to me code wise!

cartmouse (approved):
I've had a look over these latest changes and it all looks good to me code wise!

cartmouse:
Reviewed up to here, all looking good!

PR #335, opened 2022-08-04
Fix input CANCEL event from 'cancel' to 'pointercancel'.

JamesProvan-UL:
The input CANCEL event is currently using 'cancel' as the PointerEvent name.

It should be using '[pointercancel](https://developer.mozilla.org/en-US/docs/Web/API/Element/pointercancel_event)' to match the other pointer events.


JordanBerni-UL:
Looks like a logical fix to me 👍 will for @NatGreenwood-UL or @cartmouse to say they're happy with it as well before approving

cartmouse:
Unless there's any historical reason why it should be `cancel` rather than `pointercancel`, I'm happy with this!

JordanBerni-UL (approved):

TF_Tooling_Web/src/InputControllers/WebInputController.ts
```diff
@@ -119,14 +119,21 @@ export class WebInputController extends BaseInputController {
         switch (_inputData.InputType) {
             case InputType.CANCEL:
                 this.ResetScrollData();
-                let cancelEvent: PointerEvent = new PointerEvent("cancel", this.activeEventProps);
+                let cancelEvent: PointerEvent = new PointerEvent("pointercancel", this.activeEventProps);
+                let outEvent: PointerEvent = new PointerEvent("pointerout", this.activeEventProps);
+
+                if (this.lastHoveredElement !== null) {
```
GabrielHomewoodUL: I thinks this should also check that it isn't elementAtPos as well as otherwise we may get duplicate messages

GabrielHomewoodUL (approved):

PR #349, opened 2022-09-01
Port Unity Settings UI to using TouchFree's TrackingAPI implementation

NatGreenwood-UL:
- [x] Code Review
- [x] Developer Testing
- [x] QA Review
- [x] Test Plan updated
- [ ] PO Review (Optional)
- [ ] Update Release Notes


TF_Settings_and_Tooling_Unity/Assets/TouchFree/SettingsUI/Scripts/Configuration/Screens/CameraPreviewScreen.cs
```diff
@@ -107,46 +81,63 @@ void OnOverExposureValueChanged(bool state)
...
+            SetSliders(
+                maskData.Value.left,
+                maskData.Value.right,
+                maskData.Value.upper,
+                maskData.Value.lower
+            );
+        }
+
+        if (cameraReversed.HasValue) {
+
```
GabrielHomewoodUL: Unnecessary new line and the braces for the ifs aren't on new lines

NatGreenwood-UL:
Capturing here for visibility/reference; to be able to only send the change to the TrackingAPI on Slider _release_, in the commit above I added this component to each slider to go back and invoke the new `OnSliderReleased` function when they are released (couldn't find a trivial way to do this in pure code with a default Unity Slider)
image

TF_Settings_and_Tooling_Unity/Assets/TouchFree/SettingsUI/Scripts/Configuration/Screens/CameraPreviewScreen.cs
```diff
@@ -83,16 +73,16 @@ private void Update()
...
-        maskingSiderT.gameObject.SetActive(maskingAvailable);
-        maskingSiderB.gameObject.SetActive(maskingAvailable);
-        maskingDiabledWarningObject.SetActive(!maskingAvailable);
+        if (unsentMask)
+        {
+            var currentTime = DateTimeOffset.Now.ToUnixTimeMilliseconds();
+
+            if (currentTime - lastMaskTime > minMaskDelay)
+            {
+                OnSliderReleased();
```
GabrielHomewoodUL, line 83: We need to set lastMaskTime here if we update the value otherwise it will get stuck updating every frame

GabrielHomewoodUL (approved):

PR #351, opened 2022-09-05
TF-677: Screen Nudge Cursor Affordance

JordanBerni-UL:
- [ ] Code Review
- [ ] Developer Testing


TF_Tooling_Web/src/Connection/MessageReceiver.ts
```diff
@@ -248,7 +248,8 @@ export class MessageReceiver {
             let converted: TouchFreeInputAction = ConvertInputAction(action);
 
             if (converted.InputType === InputType.DOWN) {
-                CursorManager.instance.cursor.HideCloseToSwipe();
+                // TODO NOTE TODO: SHOULD THIS BE COMMENTED OUT
```
cartmouse, line 251: We should probably have this in so that the prompt disappears on a successful swipe

cartmouse (commented):
One small comment, but the rest is looking and working great!

cartmouse (approved):

PR #352, opened 2022-09-06
WebTooling: Add enable/disable to cursors

GabrielHomewoodUL:
- [x] Code Review
- [x] Developer Testing


cartmouse (commented):
All the changes look good and are working mostly as intended. 

The only thing I noticed was that the DotCursor will still render it's ring even after being disabled as the `UpdateCursor` method changes the opacity without a check for `this.enabled`

cartmouse (approved):
:shipit:

PR #357, opened 2022-09-14
Update web tooling documentation

GabrielHomewoodUL:
- [ ] Code Review
- [ ] Developer Testing

### DIFF
--- a/TF_Settings_Web/src/TouchFree/Cursors/SvgCursor.ts
+++ b/TF_Settings_Web/src/TouchFree/Cursors/SvgCursor.ts
@@ -4,13 +4,20 @@ import { MapRangeToRange } from "../Utilities";
 import { TouchlessCursor } from "./TouchlessCursor";
 
 export class SVGCursor extends TouchlessCursor {
-    xPositionAttribute: string;
-    yPositionAttribute: string;
-    cursorCanvas: SVGSVGElement;
-    cursorRing: SVGCircleElement;
-    ringSizeMultiplier: number;
+    private xPositionAttribute: string = 'cx';
+    private yPositionAttribute: string = 'cy';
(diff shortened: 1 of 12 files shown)


TF_Settings_Web/src/TouchFree/Cursors/SvgCursor.ts
```diff
@@ -4,13 +4,20 @@ import { MapRangeToRange } from "../Utilities";
 import { TouchlessCursor } from "./TouchlessCursor";
 
 export class SVGCursor extends TouchlessCursor {
-    xPositionAttribute: string;
-    yPositionAttribute: string;
-    cursorCanvas: SVGSVGElement;
-    cursorRing: SVGCircleElement;
-    ringSizeMultiplier: number;
-
-    constructor(_xPositionAttribute = "cx", _yPositionAttribute = "cy", _ringSizeMultiplier = 2, _darkCursor = false) {
```
NatGreenwood-UL, line 13: Lots of non-comment changes in here, is this an update of the SVG Cursor from the Tooling?

NatGreenwood-UL (commented):
Just one q about the changes to the SVG cursor(s), otherwise looks good

TF_Settings_Web/src/TouchFree/Cursors/SvgCursor.ts
GabrielHomewoodUL: Sorry, I ended up using this as an opportunity to remove a couple of variables and make some things private that I didn't think would be useful to have exposed. As far as I know they are not used by anyone else and we haven't included them in a release of web tooling yet

NatGreenwood-UL (approved):

PR #359, opened 2022-09-14
Feature/tf 650 camera connection events

Craig-J:
- [x] Code Review
- [x] Developer Testing
- [x] QA Review
- [x] Test Plan updated
- [x] PO Review (Optional)
- [x] Update Release Notes

## Summary

Add connection status event to TouchFree Tooling and .NET Service.

#### Service
I've sparingly made some changes that show language features I believe are helpful (which may be new to the reviewers) as well as some changes that are iterating towards improving the architecture of the overall product.

- `ClientConnectionManager` references `IConfigManager` and subscribes to `ConnectionManager.ServiceStatusChange` event through `IHandManager`.
- `ITrackingConnectionManager` now has getter and event for Service Status.
- Tidied up some of the implementation in `TrackingConnectionManager` and `HandManager`.
- Removed passthrough connection methods from `IHandManager` and replaced with a refence to `ITrackingConnectionManager`.
- Completed `IController` interface and replaced concrete `Controller` references.
- Added `Result<T>` for usage in `ITrackingDiagnosticAPI` events. `Result` is easier/safer to work with than nullable values with separate error messages.
- Modified members of `ITrackingDiagnosticAPI` to improve readability and remove implementation details.
- Refactored TrackingDiagnosticAPI implementation. This implementation will communicate relevant calls to the diagnostic API even when the device is disconnected when those calls are valid without a device connected. It now also caches config state itself and manages communicating this state to the diagnostic API when a device connects - pushing the values of "initialized" and pulling the values of "uninitialized" state. The behaviour of config state is as follows: "state set from client requests" > "state loaded from config" > "state existing on a device, retrieved when connected" > "default state value (uninitialized)". The Diagnostic API will save config state changes to file even when a device is disconnected if all values are initialized (e.g. there are none in the last state) - this behaviour is to avoid "state existing on a connected device" from getting overwritten by "loaded from config" in the case of default values.

Some further comments that I didn't do anything about:
- The chain of interfaces `IHandManager` -> `ITrackingConnectionManager` -> `IController` is curious. Including lower level interfaces via composition could be considered leaky abstraction but forwarding all of the calls on is not a reasonable solution to avoid it.
- Implementing the ability to save partial config files would allow simplifying the new implementation of Diagnostic API so that it doesn't have avoid saving to config unless all state is "initialized". It could just save only the initialized values.

#### Web Tooling

- Added `TouchFreeEvent` string enum to centralize event strings used for event listeners throughout tooling.
- Added an event dispatch from `ConnectionManager` for `TrackingServiceState` events that don't match up to a request id. These are assumed to be tracking service state changes that are pushed from the service.
- Added `AddServiceStatusListener` method to `ConnectionManager` for users to register to receive the service state dispatches mentioned above.

#### Web Settings

- Subscribe to `ServiceStatusListener` and replace polling behaviour with event. Service status will now update immediately upon device connection/disconnection.
- TouchFree Service Status is also passed through to `MaskingScreen` but nothing is done with it currently. Ideally it would update the UI in some way to show what actions cannot be performed due to device being disconnected.

## Issues Discovered/Still to fix

- Service will often crash when debugging due to incredibly long stack trace. More logging breakpoints seems to make it occur more often. Never ran into it when not debugging - seems to be performance sensitive. I suspect it's related to this issue https://github.com/sta/websocket-sharp/issues/702. Need to spend more time investigating.
- Masking screen will not read/save configuration until a device has been connected. I don't believe this is originating from the changed service Diagnostic API implementation. Does web settings/tooling prevent changes until status is connected? Same behaviour on develop - investigate separately.
- WebSocket exception usually occurs in service when refreshing web settings. `System.Net.WebSockets.WebSocketException (0x80004005): The remote party closed the WebSocket connection without completing the close handshake.` Easy to reproduce if required. Also happens on develop - should be investigated separately.
- Error when debugging web settings in debug console for `MaskingSliderDraggable`. Unsure if this is causing issues or not. Same behaviour on develop - investigate separately.
image

### DIFF
--- a/TF_Service_dotNet/TouchFree/Connections/TrackingDiagnosticAPI.cs
+++ b/TF_Service_dotNet/TouchFree/Connections/TrackingDiagnosticAPI.cs
@@ -30,6 +30,10 @@ namespace Ultraleap.TouchFree.Library.Connections
         public event Action OnTrackingServerInfoResponse;
         public event Action OnTrackingDeviceInfoResponse;
 
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <typeparam name="TData"></typeparam>
         private class ConfigurationVariable<TData>
         {
             private readonly Action<ConfigurationVariable<TData>> _onSet;
@@ -171,7 +175,19 @@ namespace Ultraleap.TouchFree.Library.Connections
                 // If nothing changed, we don't need to save - this is to avoid infinite looping caused by events
                 if (configChanged)
                 {
-                    TrackingConfigFile.SaveConfig(new TrackingConfig());
+                    TrackingConfigFile.SaveConfig(new TrackingConfig
+                    {
+                        Mask =
+                        {
+                            Left = (float)maskingData.Value.left,
+                            Right = (float)maskingData.Value.right,
+                            Lower = (float)maskingData.Value.lower,
+                            Upper = (float)maskingData.Value.upper
+                        },
+                        AllowImages = allowImages.Value,
+                        AnalyticsEnabled = analyticsEnabled.Value,
+                        CameraReversed = cameraReversed.Value
+                    });
                 }
             }
         }
--- a/TF_Service_dotNet/TouchFree/Connections/TrackingDiagnosticAPI.cs
+++ b/TF_Service_dotNet/TouchFree/Connections/TrackingDiagnosticAPI.cs
@@ -11,15 +11,13 @@ namespace Ultraleap.TouchFree.Library.Connections
     public class TrackingDiagnosticApi : ITrackingDiagnosticApi, IDisposable
     {
         private static string uri = "ws://127.0.0.1:1024/";
-        private const string minimumMaskingAPIVerison = "2.1.0";
-        private IConfigManager configManager;
+        private readonly IConfigManager configManager;
         private string trackingServiceVersion;
         private Version version;
 
         public uint? connectedDeviceID;
         public string connectedDeviceFirmware;
         public string connectedDeviceSerial;
-        public bool maskingAllowed; // TODO: Is this relevant anymore? Unused.
         
         public event Action<Result<ImageMaskData>> OnMaskingResponse;
         public event Action<Result<bool>> OnAnalyticsResponse;
@@ -155,7 +153,7 @@ namespace Ultraleap.TouchFree.Library.Connections
         private void TrySaveTrackingConfiguration()
         {
             // We don't save the tracking config until all values are initialized to prevent situations such as defaults
-            // always being written to config when a device hasn't been connected yet and there's no existing config
+            // always being written to config when a device hasn't been connected yet when there's no existing config
             if (maskingData.Initialized &&
                 allowImages.Initialized &&
                 cameraReversed.Initialized &&
@@ -220,18 +218,18 @@ namespace Ultraleap.TouchFree.Library.Connections
                 webSocket.OnMessage += OnMessage;
                 webSocket.OnOpen += (sender, e) =>
                 {
-                    Console.WriteLine("DiagnosticAPI open... ");
+                    TouchFreeLog.WriteLine("DiagnosticAPI open... ");
                     RequestGetServerInfo();
                     RequestGetDevices();
                     RequestGetVersion();
                 };
                 webSocket.OnError += (sender, e) =>
                 {
-                    Console.WriteLine("DiagnosticAPI error! " + e.Message + "\n");
+                    TouchFreeLog.ErrorWriteLine($"DiagnosticAPI error! {e.Message}\n");
                 };
                 webSocket.OnClose += (sender, e) =>
                 {
-                    Console.WriteLine("DiagnosticAPI closed. " + e.Reason);
+                    TouchFreeLog.WriteLine($"DiagnosticAPI closed. {e.Reason}");
                 };
             }
 
@@ -241,7 +239,7 @@ namespace Ultraleap.TouchFree.Library.Connections
             }
             catch (Exception ex)
             {
-                Console.WriteLine("DiagnosticAPI connection exception... " + "\n" + ex.ToString());
+                TouchFreeLog.ErrorWriteLine($"DiagnosticAPI connection exception... \n{ex}");
             }
         }
 
@@ -252,6 +250,7 @@ namespace Ultraleap.TouchFree.Library.Connections
                 // + " " is used to avoid stack overflowing by creating a new string to
                 // avoiding keeping around the event args object
                 // TODO: Different method of avoiding stack overflow exceptions?
+                // NOTE: Suspected that overflows are related to this issue https://github.com/sta/websocket-sharp/issues/702
                 newMessages.Enqueue(e.Data + " ");
             }
         }
@@ -269,7 +268,7 @@ namespace Ultraleap.TouchFree.Library.Connections
                     var payload = JsonConvert.DeserializeObject<DApiPayloadMessage<TPayload>>(_message);
                     if (payload == null)
                     {
-                        Console.WriteLine($"DiagnosticAPI - Payload for {status.ToString()} failed to deserialize: {_message}");   
+                        TouchFreeLog.WriteLine($"DiagnosticAPI - Payload for {status.ToString()} failed to deserialize: {_message}");   
                     }
                     else
                     {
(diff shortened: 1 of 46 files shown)


Craig-J:
I've marked this as ready so that review can proceed while I spend a bit more time investigating issues.

TF_Settings_Web/src/App.tsx
```diff
@@ -16,27 +16,27 @@ const App: React.FC = () => {
...
-        const controller: WebInputController = new WebInputController();
+        const requestServiceStatus = () => ConnectionManager.RequestServiceStatus((detail: ServiceStatus) => {
+            const status = detail.trackingServiceState;
+            if (status) {
+                setTfStatus(status);
+            }
+        });
 
-        const timerID = window.setInterval(updateTfStatus, 5000);
+        const updateTfStatus = (serviceStatus:TrackingServiceState) => {
```
GabrielHomewoodUL: We don't need the updateTfStatus method any more, I think you can just pass in the setTfStatus where it is used instead

TF_Settings_Web/src/Components/Pages/Camera/Masking/MaskingScreen.tsx
```diff
@@ -23,7 +23,12 @@ export type Lens = 'Left' | 'Right';
 
 const FRAME_PROCESSING_TIMEOUT = 60;
 
-const MaskingScreen = () => {
+interface MaskingScreenProps
+{
+    tfStatus:TrackingServiceState;
```
GabrielHomewoodUL: This doesn't appear to be used anywhere

TF_Settings_Web/src/TouchFree/Connection/ConnectionManager.ts
```diff
@@ -7,125 +7,136 @@ import { HandPresenceState, ServiceStatus } from "./TouchFreeServiceTypes";
 // This Class manages the connection to the Service. It provides static variables
 // for ease of use and is a Singleton to allow for easy referencing.
 export class ConnectionManager extends EventTarget {
```
GabrielHomewoodUL, line 9: Can you set the whitespace back to 2 spaces instead of 4

TF_Settings_Web/src/Components/Pages/Camera/Masking/MaskingScreen.tsx
Craig-J: Yes it's unused but I left it in because it will be needed when we update the masking screen UI based on state. Can remove for now if you'd prefer.

TF_Settings_Web/src/TouchFree/Connection/ConnectionManager.ts
Craig-J: You mean 4 instead of 2?

TF_Service_dotNet/TouchFree/Connections/TrackingDiagnosticAPI.cs
```diff
@@ -11,53 +12,184 @@ public class TrackingDiagnosticApi : ITrackingDiagnosticApi, IDisposable
     {
         private static string uri = "ws://127.0.0.1:1024/";
         private const string minimumMaskingAPIVerison = "2.1.0";
-        private readonly IConfigManager configManager;
-        public string trackingServiceVersion;
-        public Version version;
+        private IConfigManager configManager;
```
GabrielHomewoodUL: The configManager should stay readonly

TF_Service_dotNet/TouchFree/Connections/TrackingDiagnosticAPI.cs
```diff
@@ -11,53 +12,184 @@ public class TrackingDiagnosticApi : ITrackingDiagnosticApi, IDisposable
...
         public string connectedDeviceSerial;
-        public bool maskingAllowed = false;
-        public bool cameraReversed = false;
-        public bool? allowImages;
-
-        public event Action<ImageMaskData?, String> OnMaskingResponse;
-        public event Action<bool?, string> OnAnalyticsResponse;
-        public event Action<bool?, string> OnAllowImagesResponse;
-        public event Action<bool?, string> OnCameraOrientationResponse;
+        public bool maskingAllowed; // TODO: Is this relevant anymore? Unused.
```
GabrielHomewoodUL: We can remove this as we know that masking is available with the tracking supplied

TF_Service_dotNet/TouchFree/Connections/TrackingDiagnosticAPI.cs
```diff
@@ -120,24 +242,47 @@ private void Connect()
...
+                        Console.WriteLine($"DiagnosticAPI - Payload for {status.ToString()} failed to deserialize: {_message}");   
+                    }
+                    else
+                    {
+                        onSuccess(payload.payload);
+                    }
+                }
+                catch
+                {
+                    Console.WriteLine($"DiagnosticAPI - Could not parse {status.ToString()} data: {_message}");
```
GabrielHomewoodUL: These Console.WriteLine methods should use the TouchFreeLog.WriteLine instead

TF_Settings_Web/src/TouchFree/Connection/ConnectionManager.ts
GabrielHomewoodUL: Yes, got them the wrong way round

TF_Service_dotNet/TouchFree/HandManager.cs
```diff
@@ -102,55 +63,28 @@ public Hand RightHand
...
 
         public void UpdateTrackingTransform(PhysicalConfigInternal _config)
         {
             // To simplify the configuration values, positive X angles tilt the Leap towards the screen no matter how its mounted.
             // Therefore, we must convert to the real values before using them.
             // If bottom mounted, the X rotation should be negative if tilted towards the screen so we must negate the X rotation in this instance.
-            var isTopMounted = ((_config.LeapRotationD.Z > 179.9f) && (_config.LeapRotationD.Z < 180.1f));
-            float xAngleDegree = isTopMounted ? _config.LeapRotationD.X : -_config.LeapRotationD.X;
+            var isTopMounted = _config.LeapRotationD.Z is > 179.9f and < 180.1f;
+            var xAngleDegree = isTopMounted ? _config.LeapRotationD.X : -_config.LeapRotationD.X;
```
NatGreenwood-UL: Why remove explicit declaration in favor of ambiguous when the other side is a complex expression? Makes code less maintainable imho.

TF_Settings_Web/src/Components/Pages/Camera/Masking/MaskingScreen.tsx
GabrielHomewoodUL: This is causing the build to fail. Or more specifically the unused `props` passed into the MaskingScreen

TF_Service_dotNet/TouchFree/HandManager.cs
Craig-J: I was just clobbering any highlights rider did in the area. I never need these type annotations but if you find them helpful I don't really have an opinion on turning the rule off that suggests this change.

Craig-J:
Currently investigating a race condition when saving config where it can enter a loop swapping between old and new config values.

TF_Settings_Web/src/TouchFree/Connection/MessageReceiver.ts
```diff
@@ -191,7 +198,16 @@ export class MessageReceiver {
         let serviceStatus: ServiceStatus | undefined = this.serviceStatusQueue.shift();
 
         if (serviceStatus !== undefined) {
-            MessageReceiver.HandleCallbackList(serviceStatus, this.serviceStatusCallbacks);
+            const callbackResult = MessageReceiver.HandleCallbackList(serviceStatus, this.serviceStatusCallbacks);
+
+            // If callback didn't happen for known reasons, we cna be sure it's an independent status event rather than a request response
+            if (callbackResult === 'NoCallbacksFound' || callbackResult === 'CallbacksUndefined')
+            {
+                console.log(serviceStatus)
```
NatGreenwood-UL: This console log was useful for testing this but should probably be removed now?

NatGreenwood-UL (approved):

PR #360, opened 2022-09-22
Enforce update behaviour is handled to stop multiple event firing

GabrielHomewoodUL:
- [x] Code Review
- [ ] Developer Testing
- [ ] QA Review
- [ ] Update Release Notes

### DIFF
--- a/TF_Service_dotNet/TouchFree/UpdateBehaviour.cs
+++ b/TF_Service_dotNet/TouchFree/UpdateBehaviour.cs
@@ -18,13 +18,14 @@ namespace Ultraleap.TouchFree.Library
         private void SetTimer(float framerate)
         {
             updateLoop = new(1000f / framerate);
-            updateLoop.AutoReset = true;
+            updateLoop.AutoReset = false;
             updateLoop.Enabled = true;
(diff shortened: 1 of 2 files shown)


TF_Service_dotNet/TouchFree/UpdateBehaviour.cs
```diff
@@ -18,13 +18,14 @@ public UpdateBehaviour(float framerate = TargetFPS)
...
             updateLoop = new(1000f / framerate);
-            updateLoop.AutoReset = true;
+            updateLoop.AutoReset = false;
             updateLoop.Enabled = true;
             updateLoop.Elapsed += OnTimerElapsed;
         }
 
         private void OnTimerElapsed(object sender, ElapsedEventArgs e)
         {
+            updateLoop.Start();
```
Craig-J: Move the start to after the event invocation so there cannot be overlapping event invocations - it's currently a potential race condition.

GabrielHomewoodUL: I swapped this to use a lock instead so that we wouldn't get overlapping invocations

Craig-J (approved):

PR #363, opened 2022-09-23
TF-730: Fixed bug where cursor was visible after hands lost

cartmouse:
- [x] Code Review
- [x] Developer Testing
- [ ] QA Review


TF_Tooling_Web/src/Cursors/SvgCursor.ts
```diff
@@ -89,7 +89,7 @@ export class SVGCursor extends TouchlessCursor {
 
         const position = _inputAction.CursorPosition;
 
-        if (position) {
+        if (position && this.shouldShow) {
```
GabrielHomewoodUL: I could be worth just having `shouldShow` as an early return before doing anything with this method instead. We don't really need to be doing anything with the cursor if it shouldn't be there!

cartmouse: Good shout, added!

GabrielHomewoodUL (approved):

PR #364, opened 2022-09-26
TF-740: Only send hand data messages when enabled

GabrielHomewoodUL:
- [x] Code Review
- [x] Developer Testing


cartmouse (approved):
Looks good to me! Tested and working as intended

PR #366, opened 2022-09-29
Experimental/video via service

GabrielHomewoodUL:
- [x] Code Review
- [x] Developer Testing
- [ ] QA Review


TF_Service_dotNet/TouchFree/Connections/ClientConnection.cs
```diff
@@ -62,7 +62,24 @@ public void SendHandData(HandFrame _data)
...
+
+            string jsonMessage = JsonConvert.SerializeObject(message);
+
+            byte[] jsonAsBytes = Encoding.UTF8.GetBytes(jsonMessage);
+
+            Int32 dataLength = lastHandData.Count;
+
+            IEnumerable<byte> dataToSend = BitConverter.GetBytes(dataLength);
+            dataToSend = dataLength > 0 ? dataToSend.Concat(lastHandData) : dataToSend;
+            dataToSend = dataToSend.Concat(jsonAsBytes);
```
NatGreenwood-UL, line 67: If I'm reading this right, `dataToSend` gets transmitted as one blob down the websocket, which the Tooling distinguishes by it being in the form of byte data rather than string data. In that case, do we still benefit from the `CommunicationWrapper`? That JSON will never be decoded by the big switch in `MessageReciever` now

TF_Service_dotNet/TouchFree/UpdateBehaviour.cs
```diff
@@ -39,6 +51,12 @@ private void OnTimerElapsed(object sender, ElapsedEventArgs e)
                 try
                 {
                     onUpdate?.Invoke();
+                    slowUpdateIteration++;
+                    if (slowUpdateIteration == slowUpdateCount)
```
NatGreenwood-UL: Would it be good practice here to check >= instead of ==, in case of asynchronous execution (IIRC UpdateBehaviour isn't blocking so that could happen). Not really big in this case because a. the value isn't otherwise used and b. is immediately reset but just to avoid any weird concurrency issues, might be nice

GabrielHomewoodUL: I made it blocking as part of a recent bug fix so we will only ever have one instance in here at a time, but probably worth doing anyway

TF_Service_dotNet/TouchFree/Connections/ClientConnection.cs
GabrielHomewoodUL: Yeah, that would be a good change. Another thing I'd like to add to this message is having the first byte(s) be an identifier for the message type/format to allow it to be changeable or allow different formats of binary messages in future without breaking

GabrielHomewoodUL: I've removed the wrapper now and added a flag for the binary data type

TF_Service_dotNet/TouchFree/UpdateBehaviour.cs
GabrielHomewoodUL: I've made this change too

TF_Settings_Web/src/Components/Pages/Camera/Masking/MaskingScreen.tsx
```diff
@@ -135,60 +134,79 @@ const MaskingScreen = () => {
...
+                updateCanvas(
+                    buffer.slice(8, imageArraySize),
+                    canvasContextRef.current,
+                    isCamReversed.current,
+                    showOverexposed.current
+                );
+            }
 
-        isHandProcessing.current = true;
+            const handsJson = String.fromCharCode.apply(null, [...new Uint8Array(buffer, 8 + imageArraySize)]);
```
JordanBerni-UL: Don't need to spread back into an array and use apply. Can just spread into fromCharCode()

TF_Settings_Web/src/TouchFree/Connection/ServiceConnection.ts
```diff
@@ -108,6 +108,20 @@ export class ServiceConnection {
     // types based on their <ActionCode> and added to queues on the <ConnectionManager's>
     // <MessageReceiver>.
     OnMessage(_message: MessageEvent): void {
+        if ((typeof _message.data) !== 'string') {
```
JordanBerni-UL, line 131: What happens if future messages aren't of type string?

TF_Settings_Web/src/TouchFree/Connection/ServiceConnection.ts
```diff
@@ -108,6 +108,20 @@ export class ServiceConnection {
...
+            
+            const bufferBlob = _message.data as Blob;
+            new Response(bufferBlob)
+                .arrayBuffer()
+                .then((buffer) => {
+                    const binaryDataType = new Int32Array(buffer, 0, 4)[0];
+                    if (binaryDataType === 1) {
```
JordanBerni-UL: Could this be an enum

TF_Settings_Web/src/Components/Pages/Camera/Masking/MaskingScreen.tsx
GabrielHomewoodUL: Updated this

TF_Settings_Web/src/Components/Pages/Camera/CameraSetupScreen.tsx
```diff
@@ -86,13 +54,6 @@ const CameraSetupScreen = () => {
                     text="Mask areas of your cameras vision from reflections and harsh areas of light"
                     onClick={() => navigate('masking')}
                 />
-                <Alert
-                    show={trackingServiceStatus === 'closed'}
-                    cssWidth={'60%'}
-                    text="Failed to connect to tracking service: masking unavailable"
```
JordanBerni-UL, line 92: Should we still show this but when we can't connect to service?

TF_Settings_Web/src/Components/Pages/Camera/Masking/MaskingScreen.tsx
```diff
@@ -63,41 +59,22 @@ const MaskingScreen: React.FC = () => {
     };
 
     // ===== Refs =====
-    const canvasRef = useRef<HTMLCanvasElement>(null);
-    const canvasContextRef = useRef<WebGLRenderingContext | null>(null);
-    const successfullySubscribed = useRef<boolean>(false);
+    const camFeedRef = useRef<HTMLDivElement>(null);
     const frameTimeoutRef = useRef<number>();
```
JordanBerni-UL: Can get rid of this it seems

TF_Settings_Web/src/Components/Pages/Camera/Masking/MaskingScreen.tsx
```diff
@@ -135,60 +112,44 @@ const MaskingScreen: React.FC = () => {
...
-            }, FRAME_PROCESSING_TIMEOUT);
-        } else if (!successfullySubscribed.current) {
-            socket.send(JSON.stringify({ type: 'SubscribeImageStreaming' }));
-        }
-    };
+        isHandProcessing.current = true;
 
-    const handleTFInput = (evt: CustomEvent<HandFrame>): void => {
-        if (isHandProcessing.current || !successfullySubscribed.current) return;
+        try {
```
JordanBerni-UL: Can we reduce this try catch block? Only wrap what we need to and can do an early return if we catch an error

TF_Settings_Web/src/TouchFree/Connection/MessageReceiver.ts
```diff
@@ -56,7 +56,7 @@ export class MessageReceiver {
 
     // Variable: latestHandDataItem
     // The latest <HandFrame> that has been received from the Service.
-    latestHandDataItem: any = undefined;
+    latestHandDataItem: ArrayBuffer | undefined = undefined;
```
JordanBerni-UL: Can make this optional instead of using |undefined
`latestHandDataItem?: ArrayBuffer;`

JordanBerni-UL: But it's good it's typed and not any

TF_Settings_Web/src/TouchFree/Connection/MessageReceiver.ts
```diff
@@ -290,9 +290,12 @@ export class MessageReceiver {
         let handFrame = this.latestHandDataItem;
```
JordanBerni-UL: can be const (this is common in the tooling and adding linting will solve this issue)

TF_Settings_Web/src/TouchFree/Connection/MessageReceiver.ts
```diff
@@ -290,9 +290,12 @@ export class MessageReceiver {
...
 
         if (handFrame !== undefined) {
+            this.latestHandDataItem = undefined;
             // Wrapping the function in a timeout of 0 seconds allows the dispatch to be asynchronous
             setTimeout(() => {
-                HandDataManager.HandleHandFrame(handFrame);
+                if (handFrame !== undefined) {
```
JordanBerni-UL: can check if(!handFrame)

TF_Settings_Web/src/TouchFree/Plugins/HandDataManager.ts
```diff
@@ -26,10 +24,10 @@ export class HandDataManager extends EventTarget {
     // Function: HandleHandFrame
     // Called by the <messageReceiver> to relay a <HandFrame> that has been received to any
     // listeners of <TransmitHandData>.
-    public static HandleHandFrame(_data: HandFrame): void {
+    public static HandleHandFrame(_data: ArrayBuffer): void {
         const currentTimeStamp = Date.now();
-        if (!HandDataManager.lastFrame || HandDataManager.lastFrame + 100 < currentTimeStamp ) {
-            let rawHandsEvent: CustomEvent<HandFrame> = new CustomEvent<HandFrame>(
+        if (!HandDataManager.lastFrame || HandDataManager.lastFrame + 16 < currentTimeStamp ) {
```
JordanBerni-UL: What is this magic 16 for?

TF_Tooling_Web/src/Connection/MessageReceiver.ts
```diff
@@ -56,7 +56,7 @@ export class MessageReceiver {
 
     // Variable: latestHandDataItem
     // The latest <HandFrame> that has been received from the Service.
-    latestHandDataItem: any = undefined;
+    latestHandDataItem: ArrayBuffer | undefined = undefined;
```
JordanBerni-UL: Same comments as other tooling (won't duplicate rest of comments)

TF_Settings_Web/src/Components/Pages/Camera/Masking/createCameraData.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    // Set black pixels to remove flashing camera bytes
+    const startOffset = isCameraReversed ? 0 : (lensHeight - 1) * width;
+    buf32.fill(0xff000000, startOffset, startOffset + width);
+
+    updateCameraRender(buf8, width / 2, lensHeight / 2, handData);
+    renderScene();
+};
+
+const processScreen = (
```
JordanBerni-UL: Weird name (did I call it that?) - maybe processCameraFrame?

TF_Settings_Web/src/Components/Pages/Camera/Masking/createHandData.ts
```diff
@@ -0,0 +1,95 @@
...
+    return {
+        tip: translateToCoordinate(finger?.Bones[tipJointIndex]?.NextJoint),
+        knuckle: translateToCoordinate(finger?.Bones[knuckleJointIndex]?.PrevJoint),
+    };
+};
+
+const zMax = 0.1;
+
+const translateToCoordinate = (coordinate?: Vector) => {
+    if (coordinate === undefined) return new Vector3(-2, -2, -2);
```
JordanBerni-UL: if(!coordinate)

TF_Settings_Web/src/Components/Pages/Camera/Masking/createHandData.ts
```diff
@@ -0,0 +1,95 @@
...
+    const finger = fingers.find((f: RawFinger) => f.Type == fingerType);
+    return {
+        tip: translateToCoordinate(finger?.Bones[tipJointIndex]?.NextJoint),
+        knuckle: translateToCoordinate(finger?.Bones[knuckleJointIndex]?.PrevJoint),
+    };
+};
+
+const zMax = 0.1;
+
+const translateToCoordinate = (coordinate?: Vector) => {
```
JordanBerni-UL: Add a return type

TF_Settings_Web/src/Components/Pages/Camera/Masking/sceneRendering.ts
```diff
@@ -0,0 +1,299 @@
...
+import cssVariables from 'Styles/variables.module.scss';
+
+import { FingerType } from 'TouchFree/TouchFreeToolingTypes';
+import { MapRangeToRange } from 'TouchFree/Utilities';
+
+import { FingerData, HandData, HandState } from './createHandData';
+
+export type BasicMesh = Mesh<BufferGeometry, MeshBasicMaterial>;
+
+interface FingerMesh {
```
JordanBerni-UL: Calling it a mesh is a bit misleading but not really sure of a better name
FingerMeshInfo?

TF_Settings_Web/src/Components/Pages/Camera/Masking/sceneRendering.ts
```diff
@@ -0,0 +1,299 @@
...
+    meshUpdateData.positions = meshUpdateData.positions.concat(...wristData.positions);
+
+    updateCircleMeshes(handMesh.circleMeshes, meshUpdateData, scale);
+};
+
+const updateFingerMesh = (
+    fingerMesh: FingerMesh,
+    finger?: FingerData,
+    nextKnucklePos?: Vector3,
+    scale?: number
```
JordanBerni-UL: scale is either 1 or comes from MapRangeToRange so doesn't need to be optional

TF_Settings_Web/src/Components/Pages/Camera/Masking/sceneRendering.ts
```diff
@@ -0,0 +1,299 @@
...
+    }
+
+    return data;
+};
+
+const updateWristMesh = (
+    wristMesh: WristMesh,
+    wrist?: Vector3,
+    thumbKnucklePos?: Vector3,
+    scale?: number
```
JordanBerni-UL: same as above

JordanBerni-UL (changes requested):

TF_Settings_Web/package.json
```diff
@@ -8,6 +8,7 @@
         "react-canvas-confetti": "^1.3.0",
         "react-circular-progressbar": "^2.1.0",
         "react-dom": "^17.0.2",
+        "three": "^0.143.0",
         "twgl.js": "^4.24.0",
```
JordanBerni-UL: This can probably go

TF_Service_dotNet/TouchFree/Connections/ClientConnection.cs
```diff
@@ -55,7 +55,22 @@ public void SendHandData(HandFrame _data)
...
+            IEnumerable<byte> binaryMessageType = BitConverter.GetBytes((int)BinaryMessageType.Hand_Data);
+            var dataToSend = binaryMessageType.Concat(BitConverter.GetBytes(dataLength));
+            dataToSend = dataLength > 0 ? dataToSend.Concat(lastHandData) : dataToSend;
+            dataToSend = dataToSend.Concat(jsonAsBytes);
+
+            Socket.SendAsync(
+                dataToSend.ToArray(),
+                WebSocketMessageType.Binary,
+                true,
+                CancellationToken.None);
```
Craig-J, line 73: I suspect we can make this bit a lot faster and reduce allocations by using a binary writer or something from `System.Memory`. Spent time trying to add benchmarking to test this theory today and ended up wasting hours due to the service stack overflowing in the debugger. Will approve this code and come back to it another time when above has been resolved.

TF_Service_dotNet/TouchFree/HandManager.cs
```diff
@@ -58,21 +60,42 @@ public class HandManager : IHandManager
...
+        }
+
+        public bool TrackingServiceConnected()
+        {
+            return ConnectionManager?.Controller?.IsServiceConnected ?? false;
+        }
+
+        public bool CameraConnected()
+        {
+            return ConnectionManager?.Controller?.Devices?.ActiveDevice != null;
```
Craig-J: Unreferenced (should be private at least?)

JordanBerni-UL:
Happy with the UI side of the changes

TF_Service_dotNet/TouchFree/Connections/ClientConnection.cs
GabrielHomewoodUL: This isn't actually doing many allocations. It is doing most of the work on the `.ToArray` call but otherwise it isn't actually performing the work of the `.Concat` calls until the `.ToArray` call

TF_Service_dotNet/TouchFree/HandManager.cs
GabrielHomewoodUL: Removed these as they aren't used

TF_Service_dotNet/TouchFree/Connections/ClientConnection.cs
Craig-J: Each concat allocates an iterator, ToArray will always allocate a new array, JsonConvert will allocate a bunch of stuff that could be reused between calls instead too. It's not that much each call, not particularly worried about these. Main thing I wanted to test is I remember BitConverter being quite slow compared to alternatives.

JordanBerni-UL (approved):

Craig-J:
Performance is much better than before.
First time going to camera masking screen the image render is broken. Have to go back out and in again.

PR #369, opened 2022-10-10
TF-758: CI Checks web settings source trees are identical

JordanBerni-UL:
- [x] Code Review
- [ ] Developer Testing

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -10,11 +10,11 @@ variables:
 
   BS_EXTENSION_NAME: ultouchfree
 
-  TF_TFU_VERSION: "1.1.0"
-  TF_TFW_VERSION: "1.1.1"
-  TOUCHFREE_VERSION: "2.3.1"
+  TF_TFU_VERSION: '1.1.0'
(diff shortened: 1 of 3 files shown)


GabrielHomewoodUL (approved):

Craig-J:
This works but doesn't produce an artefact to fix issues. Could we use git diff to produce a patch somehow?
See formatting job [here ](https://github.com/ultraleap/UnityPlugin/blob/develop/.gitlab-ci.yml) as an example (ignore the copyright headers bit) of how to use git diff to produce a patch

Craig-J:
Is the current workflow generally to modify the tooling in web settings? If that's what we usually do then you could copy over the tooling source from web settings and use git diff to produce a patch.

JordanBerni-UL:
> Is the current workflow generally to modify the tooling in web settings? If that's what we usually do then you could copy over the tooling source from web settings and use git diff to produce a patch.

I don't think we can assume which source is the correct one as changes could come from either. A failing pipeline job should be enough to remind the author to add the missing changes to the other source. This should be resolved in the future when there is one true source

PR #374, opened 2022-10-26
TF-774: Versioning Improvements

GabrielHomewoodUL:
- [x] Code Review
- [x] Developer Testing
- [ ] QA Review
- [ ] Test Plan updated
- [ ] PO Review (Optional)
- [ ] Update Release Notes


TF_Service_dotNet/TouchFree/Configuration/VersionManager.cs
```diff
@@ -0,0 +1,25 @@
...
+            get
+            {
+                if (string.IsNullOrEmpty(_version))
+                {
+                    _version = Assembly.GetExecutingAssembly().GetCustomAttribute<AssemblyInformationalVersionAttribute>().InformationalVersion;
+                }
+
+                return _version;
+            }
+        }
```
Craig-J: Could replace with `_version ??= Assembly.GetExecutingAssembly().GetCustomAttribute<AssemblyInformationalVersionAttribute>()!.InformationalVersion;`
Set `_version` default to null.
Note I added a null suppression operator to fix warning that the attribute could be missing.

.gitlab-ci.yml
```diff
@@ -173,8 +185,14 @@ TF_Service::dotNet::ARMLinux::build:
   when: always
   image: registry.ultrahaptics.com/xr/infrastucture/dotnet6-toolchain:latest
   script:
-    - dotnet publish -c Release -r linux-arm64 "$CI_PROJECT_DIR/TF_Service_dotNet/TouchFree/TouchFreeLib.csproj"
-    - dotnet publish -c Release -r linux-arm64 "$CI_PROJECT_DIR/TF_Service_dotNet/TouchFree_Service/TouchFree_Service.csproj"
+    - if ($CI_COMMIT_BRANCH -Match 'release/.*') {
+        $TOUCHFREE_SUFFIX = 'beta'
+      } else {
+        $TOUCHFREE_SUFFIX = 'beta-' + ${env:CI_COMMIT_SHORT_SHA}
```
Craig-J: Use `+` rather than `-` for build metadata as per semver spec. (Same for above 2 builds if they don't include the `+`.

cartmouse (approved):
All works - looks good to me!

TF_Service_dotNet/TouchFree/Configuration/VersionManager.cs
GabrielHomewoodUL: I've updated this to handle the attribute not existing instead

.gitlab-ci.yml
GabrielHomewoodUL: Done, and put a dev prefix for dev versions

PR #381, opened 2022-11-02
Simplify Initialisation of TouchFree Tooling

NatGreenwood-UL:
- [x] Code Review
- [x] Developer Testing
- [ ] QA Review
- [ ] Test Plan updated
- [ ] PO Review (Optional)
- [ ] Update Release Notes


TF_Tooling_Web/src/Cursors/SvgCursor.ts
```diff
@@ -3,6 +3,11 @@ import { InputType, TouchFreeInputAction } from "../TouchFreeToolingTypes";
 import { MapRangeToRange } from "../Utilities";
 import { TouchlessCursor } from "./TouchlessCursor";
 
+export interface SVGCursorParams {
+    ringSizeMultiplier: number;
+    darkCursor: boolean;
```
Craig-J: Make these optional, `FromParamObj` will pass undefined on and that will use default values in the constructor if they're not included.

Side note: constructor arguments are missing types can you add them too?

TF_Tooling_Web/src/TouchFree.ts
```diff
@@ -0,0 +1,24 @@
+import { ConnectionManager } from "./Connection/ConnectionManager";
+import { SVGCursor, SVGCursorParams } from "./Cursors/SvgCursor";
+import { TouchlessCursor } from "./Cursors/TouchlessCursor";
+import { WebInputController } from "./InputControllers/WebInputController";
+
+var InputController: WebInputController;
+export var CurrentCursor : TouchlessCursor;
+
+export function Init(_cursor: TouchlessCursor): void;
+export function Init(_svgParams?: SVGCursorParams): void;
```
Craig-J: Overloads are unnecessary

TF_Tooling_Web/src/TouchFree.ts
```diff
@@ -0,0 +1,24 @@
...
+
+    if (_cursorOrParams instanceof TouchlessCursor) {
+        CurrentCursor = _cursorOrParams;
+    }
+    else if (_cursorOrParams !== undefined) {
+        CurrentCursor = SVGCursor.FromParamObj(_cursorOrParams);
+    } else {
+        CurrentCursor = new SVGCursor();
+    }
+}
```
Craig-J: I would suggest having the signature be an optional interface with optional properties so we can extend it in future

```suggestion
type CursorInit = TouchlessCursor | SVGCursorParams;

interface InitParams {
    Cursor?: CursorInit
}

export function Init(params?:InitParams): void {
    ConnectionManager.init();

    InputController = new WebInputController();

    if (params?.Cursor === undefined)
    {
        CurrentCursor = new SVGCursor();
    }
    else if (params.Cursor instanceof TouchlessCursor) {
        CurrentCursor = params.Cursor;
    }
    else
    {
        CurrentCursor = SVGCursor.FromParamObj(params.Cursor);
    }
}
```

TF_Tooling_Web/src/TouchFree.ts
```diff
@@ -0,0 +1,29 @@
...
+import { SVGCursor, SVGCursorParams } from "./Cursors/SvgCursor";
+import { TouchlessCursor } from "./Cursors/TouchlessCursor";
+import { WebInputController } from "./InputControllers/WebInputController";
+
+var InputController: WebInputController;
+export var CurrentCursor: TouchlessCursor;
+
+export function Init(_cursor: TouchlessCursor): void;
+export function Init(_svgParams?: SVGCursorParams): void;
+export function Init(_cursorOrParams?: SVGCursorParams | TouchlessCursor): void {
```
GabrielHomewoodUL: I feel like we could just have an object that handles these two cases e.g.
`{ cursor: TouchlessCursor, svgCursorParams: SVGCursorParams, }`
And as long as we make it clear that you can create this shape object then it would be clearer and be more extensible in future.

JordanBerni-UL: I agree with Gabe - there should be an optional options interface you can pass in the init function?

export function Init(options?: TouchFreeInitOptions)

interface TouchFreeInitOptions {
     cursor?: TouchlessCursor
}

Instead of passing svgCursor params could the user not create an SVGCursor (using the params) then pass it in via the TouchFreeInitOptions?

TF_Tooling_Web/src/TouchFree.ts
```diff
@@ -0,0 +1,29 @@
...
+    ConnectionManager.AddConnectionListener(() => {
+        InputController = new WebInputController();
+
+        if (_cursorOrParams instanceof TouchlessCursor) {
+            console.log("arg was a Cursor");
+            CurrentCursor = _cursorOrParams;
+        }
+        else if (_cursorOrParams !== undefined) {
+            console.log("arg was cursor params");
+            CurrentCursor = SVGCursor.FromParamObj(_cursorOrParams);
```
JordanBerni-UL: Could we not just create a new SVGCursor with the params extracted from the object? Instead of needing FromParamObj?

TF_Tooling_Web/src/TouchFree.ts
Craig-J: See my suggestion here https://github.com/ultraleap/TouchFree/pull/381#discussion_r1012056817 (not sure why it's showing outdated now)

JordanBerni-UL: I don't think the SVGCursorParams are needed? Can't they make a new SVG cursor then pass it in? Or is that a weird / hidden workflow

TF_Tooling_Web/src/TouchFree.ts
Craig-J: Agree with this - the ideal solution is we change the constructor to take an interface/union of config but that's breaking so we should come back to that later

TF_Tooling_Web/src/TouchFree.ts
NatGreenwood-UL: What's the benefit of using an interface containing the type here rather than just the type? I suppose it allows us the freedom to add additional params without it being a breaking change?

JordanBerni-UL: Also if we end up with a lot of init options this is cleaner and easier to understand by keeping the actual Init function small (rather than loads of parameters)

TF_Tooling_Web/src/TouchFree.ts
NatGreenwood-UL: I'd designed for three approaches to the Init:

1. Just make it work, I don't care for the details
2. This ^ but I wanna tweak something
3. I want full control, use this thing I made

2 & 3 could both be fulfilled by just the Cursor passing, only difference is it requires a user to 
understand that the SVG cursor is the default, & then the step of them making said cursor beforehand.

All this said, I'm starting to fear that all of this is actually redundant; I'm not sure the case of passing in a cursor *will ever work*. Cursors won't be able to initialise successfully without the ConnectionManager already being instanced... so you need some of the function to have run before you instace the thing you're passing _into_ that function. 
We could replace this with a factory function you provide maybe? As a "once you've initialised the connection, use this function to make a Cursor"?

TF_Tooling_Web/src/TouchFree.ts
NatGreenwood-UL: We could but then we're basically duplicating the code in FromParamObj everywhere we do this conversion, it's just a tiny convenience function whilst we can't make the breaking change described above. Don't see a reason _not_ to have a tiny function that saves me typing out the pulling out of all the members of an object as params.

TF_Tooling_Web/src/TouchFree.ts
JordanBerni-UL: Maybe then we only support 1? And if you want a non-default cursor you have to do it the long way?

NatGreenwood-UL: Maybe we support 1 & 3 but we have two seperate functions:

`InitAll` does what `Init()` currently does; gets you a Connection, Inputs & a Cursor
`InitBase` only initialises the bits you don't need to think about (Connection & Inputs) and leaves you to manage your cursor(s) as you please?

NatGreenwood-UL: Alternatively maybe it's one function with a bool dictating whether it makes a cursor for you, which defaults to true?

JordanBerni-UL: ^ personally I like that option the most.
Or could make the bool inside the InitParams interface and have DefaultInitParms? - but that's only useful if we envision adding lots more options

TF_Tooling_Web/src/Cursors/SvgCursor.ts
NatGreenwood-UL: This is done in the latest commit

TF_Tooling_Web/src/TouchFree.ts
NatGreenwood-UL: These are gone now, we have one function sig with an interface as discussed elsewhere

TF_Tooling_Web/src/TouchFree.ts
NatGreenwood-UL: This is done now, though we now only have one option (cursor / no)

TF_Tooling_Web/src/TouchFree.ts
NatGreenwood-UL: The change I described in my previous comment is now implemented 👍

NatGreenwood-UL:
The build is currently failing because the tooling versions are out of synch, I'll fix this once we're happy with the implementation in one place 👍

TF_Tooling_Web/src/Cursors/SvgCursor.ts
```diff
@@ -3,6 +3,11 @@ import { InputType, TouchFreeInputAction } from "../TouchFreeToolingTypes";
 import { MapRangeToRange } from "../Utilities";
 import { TouchlessCursor } from "./TouchlessCursor";
 
+export interface SVGCursorParams {
```
JordanBerni-UL: Doesn't look this is used anywhere now?

TF_Tooling_Web/src/Cursors/SvgCursor.ts
```diff
@@ -80,6 +85,10 @@ export class SVGCursor extends TouchlessCursor {
         ConnectionManager.instance.addEventListener('HandsLost', this.HideCursor.bind(this));
     }
 
+    static FromParamObj(params: SVGCursorParams) : SVGCursor {
```
JordanBerni-UL: Doesn't look this is used anywhere now?

TF_Tooling_Web/src/TouchFree.ts
```diff
@@ -0,0 +1,29 @@
...
+    initialiseCursor?: boolean;
+}
+
+export function Init(_tfInitParams?: TfInitParams ): void {
+    ConnectionManager.init();
+
+    ConnectionManager.AddConnectionListener(() => {
+        InputController = new WebInputController();
+
+        if (_tfInitParams === undefined) {
```
JordanBerni-UL, line 19: IMO `if(!_tfInitParams)` looks cleaner but not a blocker

TF_Tooling_Web/src/TouchFree.ts
```diff
@@ -0,0 +1,29 @@
...
+            CurrentCursor = new SVGCursor();
+        } else {
+            if (_tfInitParams.initialiseCursor === undefined) {
```
JordanBerni-UL: Could combine this into one check?

TF_Tooling_Web/src/TouchFree.ts
NatGreenwood-UL: `!var` can return true in a wide array of situations because of JS type enforcement (e.g. it'll also return true if it's `null` or if it's `false`) where checking explicitly only covers `undefined` which is what I want here

TF_Tooling_Web/src/TouchFree.ts
NatGreenwood-UL: This if with the one above? If we do that the else if following this if would be messy because of not knowing what's there. I think you'd need to basically do this:

```
if (_tfInitParams === undefined || _tfInitParams.initialiseCursor === undefined) {
    CurrentCursor = new SVGCursor();
} else if (_tfInitParams !== undefined && _tfInitParams !== undefined && _tfInitParams.initialiseCursor) {
    CurrentCursor = new SVGCursor();
}
```

which I like way less

TF_Tooling_Web/src/TouchFree.ts
JordanBerni-UL: This case it's only defined or undefined so a falsey check is fine imo (agreed if it you're checking something that can be true/false/undefined)

TF_Tooling_Web/src/TouchFree.ts
JordanBerni-UL: No I meant with the one after it (initialiseCursor undefined and true check) - that is a monstrosity ^

JordanBerni-UL:
Changes look good to me - didn't want to resolve others' comments but done all of mine :)

TF_Tooling_Web/src/TouchFree.ts
Craig-J: I don't see the point in having boolean yes/no rather than an interface for params required to init. My thinking was in future the constructor for `SVGCursor` takes the params interface instead of individual parameters too.

With the boolean instead of interface we lose the flexibility of initializing with custom values. I'd quite like to see the pattern of constructor + single parameter init interface so users can pass around "stuff required to create thing" easily. In the current situation where we aren't breaking any constructors I'm not really bothered either way though.

NatGreenwood-UL: I think what we have here is a good balance between "an easy way to initialise touchfree" and "I want control of a cursor", wherein if you want control, you get full control.

Honestly the "CurrentCursor" reference here is the bit I have issue with; noone benefits from the Tooling having control of a cursor instance, it's only there in the case that you want to change the cursor you got defaulted. Ideally in most cases people will use the version that doesn't init a cursor and then init their own, and customise the cursors seperately.

NatGreenwood-UL: > we lose the flexibility of initializing with custom values

We aren't removing this flexibility, we're just pushing responsibility for initialising the custom parts that don't benefit from being controlled by tooling onto users. This gives us two scenarios:

1. You want it easy & fast, you get the default cursor
2. You want a custom cursor, so you make your own cursor (which is still _very_ simple)

I'm not arguing against the _approach_ of initialisation in the way you're recommending Craig, just that *Cursors specifically* being initialised by the Tooling Library isn't actually helpful, when it's so easy for user to manage their own.

GabrielHomewoodUL (approved):

PR #382, opened 2022-11-08
Feature/tf 605 top level events

Craig-J:
## Summary

Adds top level event registering functions in a `TouchFree` object. These functions support all events defined in `TouchFreeEventSignatures`. The register function returns a convenience `EventHandle` object that can be used to unregister the event without needing to remember the callback.

The previous enum introduced for events is replaced with a union of string literal types - note this is not breaking as the enum has not been included in any releases.

## Details

The motivation behind the current design for this PR can be split into 2 goals:

- Create an event API for TouchFree that makes discovering and consuming events easy. The API should be as helpful as possible at preventing errors in user code.
- The implementation of the API must be easy to extend with more events - it will ideally be a compile error to add events that are not properly handled in the API implementation.

### Discoverable Events & Helpful API

All of the event names and signatures are defined by the interface `TouchFreeEventSignatures`. Each new function signature added to this interface will become an event with the signature of that function. `TouchFreeEvent` is a string literal union type generated by extracting the names of the events from `TouchFreeEventSignatures`. Note that the interface is never actually implemented, it is use as a lookup for event names and function signatures.

The top level functions added are `RegisterEventCallback` and `DispatchEvent`. Both of these functions make use of generics and type indexing of `TouchFreeEventSignature` to match up the passed event string with the correct callback signature. Some example call sites of `RegisterEvent` are shown below:

```ts
RegisterEvent("OnConnected", () => {}); // ok
RegisterEvent("OnConnected", (i) => {}); // error - too many args in callback
RegisterEvent("OnTrackingServiceStateChange", (state) => {}); // ok
RegisterEvent("OnTrackingServiceStateChange", () => {}); // ok - extra arg in callback signature ignored
RegisterEvent("OnConn", () => {}); // error - OnConn is not a valid event
```

`DispatchEvent` is a variadic function which has an argument for the event to dispatch along + argument (if any for the matching event) to pass to the callbacks. This is achieved with the [rest parameter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters) syntax.

The above reasons ensure that events (and their callback function types) are discoverable from autocomplete of `RegisterEventCallback` or from looking at the list of defined events `TouchFreeEventSignatures`. Usage is easier/safer than `addEventListener` as the user doesn't need to know which object is the `EventTarget` and the compiler will ensure the callback signature is appropriate with helpful type errors.

### Maintainable Implementation

The implementation of `RegisterEventCallback` uses two mapped type objects for lookup - `EventTargets` and `EventListeners`. These lookups map all of the signatures in `TouchFreeEventSignatures` where it is a compile error both to add a mapping for an event that does not exist and to be missing a mapping for one of the signatures.

The event strings used by the top level functions, `TouchFreeEvent`, is generated from the interface `TouchFreeEventSignatures`. This makes it impossible to be missing an event with a defined signature or the reverse. It's easy to add a new event - simply add the name and signature to `TouchFreeEventSignatures` and then the compiler will show an error for the `EventTargets` and `EventListeners` lookups, prompting the programmer to handle the new event.

## Contributor Tasks

- [x] Developer testing

- [x] update the release notes on the issue
- [x] make sure the fix version on the issue is set correctly

## Reviewer Tasks

- [x] Code reviewed
- [x] Non-code assets reviewed
- [x] Checked and agree with release testing considerations added to PR for the next release

## Closes JIRA Issue

_If this PR closes any JIRA issues list them below in the form `Closes PROJECT-#`_

Closes TF-605


cartmouse:
Looks great! I really like the implementation - easy to use and understand. Seems simple enough to maintain too.

Only comment is that there are still a few leftover imports that are no longer used. I won't put a comment per case because it's most of the files that have been changed in the tooling and that just seems like spam.

cartmouse (approved):

PR #384, opened 2022-11-15
Feature/tf 685 decrease frame timeout

GabrielHomewoodUL:


cartmouse (approved):
All working as intended 👍

PR #395, opened 2022-11-23
Disable tag pipelines

Craig-J:
## Summary

Disables tag pipelines as they cause some CI script failures and we don't use them.
Note that there's no way to test this until a tag is created after being merged.


Craig-J:
Closing as we've decided to fix instead

PR #413, opened 2023-01-18
Add package lock to repo

Craig-J:
## Summary

Add npm package lock so that package resolution is consistent between machines and changes to resolved dependencies are tracked.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -76,8 +76,6 @@ Documentation/Tooling for Unity/**
 #  Testing  #
 # ========= #
 
-**/package-lock.json
-**/packages-lock.json
 **/usage.txt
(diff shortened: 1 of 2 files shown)


JordanBerni-UL (approved):

JordanBerni-UL:
Do you also need to commit the package-lock in sub folders (e.g. web settings)?

Craig-J:
> Do you also need to commit the package-lock in sub folders (e.g. web settings)?

Doesn't seem so. Initially I did `npm i` in the root and it only generated one at the root.
Just tested doing `npm i` in a subfolder in a clean install and it still only generated one in the root. Guessing it's workspace behaviour.

PR #426, opened 2023-01-26
TF-962: Interaction Zone Events

cartmouse:
## Summary

Add events for entering and leaving interaction zone. Closes [TF-962](https://ultrahaptics.atlassian.net/browse/TF-962)

### Tests Added

Tests added for registering callbacks against `TouchFree Events`.

## Contributor Tasks

- [ ] ~XDR review (optional depending on work type)~
    - [x] [TouchFree BrightSign](/ultraleap/touchfree/blob/-/CHANGELOG-brightsign.md)
    - [x] [TouchFree Web Tooling](/ultraleap/touchfree/blob/-/TF_Tooling_Web/CHANGELOG.md)

## Reviewer Tasks

- [x] Developer testing
- [x] Code reviewed
- [x] Documentation reviewed - includes checking documentation requirements are met and not missing e.g., public API is commented

[TF-962]: https://ultrahaptics.atlassian.net/browse/TF-962?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


TF_Tooling_Web/src/Cursors/SvgCursor.ts
```diff
@@ -78,8 +79,10 @@ export class SVGCursor extends TouchlessCursor {
 
         this.ringSizeMultiplier = ringSizeMultiplier;
 
-        ConnectionManager.instance.addEventListener('HandFound', this.ShowCursor.bind(this));
-        ConnectionManager.instance.addEventListener('HandsLost', this.HideCursor.bind(this));
+        TouchFree.RegisterEventCallback('HandFound', this.ShowCursor.bind(this));
+        TouchFree.RegisterEventCallback('HandsLost', this.HideCursor.bind(this));
+        TouchFree.RegisterEventCallback('HandEntered', this.ShowCursor.bind(this));
+        TouchFree.RegisterEventCallback('HandExited', this.HideCursor.bind(this));
```
GabrielHomewoodUL, line 85: Could you add these to the DotCursor too?

cartmouse: Added!

CHANGELOG-windows.md
```diff
@@ -22,7 +24,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Service: `HandEntered` & `HandExited` events, sent when the active hand enters and exits the interaction zone (if enabled) respectively
```
Craig-J, line 29: 🤷 Do we want to track this change? Is it user visible/impactful? Conscious that if we don't a lot of service changes will be undocumented though...

TF_Service_dotNet/TouchFree/InteractionZoneEvent.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+
+namespace Ultraleap.TouchFree.Library
+{
+
+    [Serializable]
+    public struct InteractionZoneEvent
```
Craig-J: 🤷 Can be shorter declaration as a record
```cs
[Serializable]
public record struct InteractionZoneEvent(InteractionZoneState state);
```

TF_Tooling_Web/src/Connection/TouchFreeServiceTypes.ts
```diff
@@ -71,6 +73,15 @@ export enum HandPresenceState {
     PROCESSED,
 }
 
+// Enum: InteractionZoneState
+// HAND_ENTERED - Sent when the "active" hand enters the interaction zone
+// HAND_EXITED - Sent when the "active" hand leaves the interaction zone
+export enum InteractionZoneState {
```
Craig-J, line 82: 🤷 Consider using two separate union types for this instead of an enum. I know it's a copy of another implementation but I don't think it's a great way to do things.
i.e.
```ts
export type InteractionZoneState = "HAND_ENTERED" | "HAND_EXITED";
export type InteractionZoneEventStatus = "PROCESSED" | "UNPROCESSED";
```
The second one could be a general purpose type instead of specific to interaction zone for other usages of this pattern.

TF_Tooling_Web/src/Connection/TouchFreeServiceTypes.ts
```diff
@@ -91,6 +102,16 @@ export class HandPresenceEvent {
     }
 }
 
+// Class: InteractionZoneEvent
+// This data structure is used to receive interaction zone requests
+export class InteractionZoneEvent {
```
Craig-J: 🤷 Consider interface instead. Alternatively the code can at least be shortened:
```ts
export class InteractionZoneEvent {
    constructor(public state: InteractionZoneState) {}
}
```

TF_Tooling_Web/src/TouchFree.ts
```diff
@@ -199,6 +213,12 @@ const EventImplementations: () => EventImpls = () =>
 //
 // TransmitInputAction: (inputAction: TouchFreeInputAction) => void;
 // Event dispatched from the <InputActionManager> to each registered Plugin
+//
+// HandEntered: () => void;
```
Craig-J, line 217: 👍 Remembered to comment this!

TF_Tooling_Web/src/Connection/TouchFreeServiceTypes.ts
cartmouse: Only issue with this is that the service uses an enum for the state so it comes through as a number. The tooling would then have to do a translation which feels like it'd just be a bit of bloat. However, I do like the type implementation.

Instead, we could do this:
```ts
export type EventStatus = "PROCESSED" | "UNPROCESSED";
export enum InteractionZoneState {
    HAND_ENTERED,
    HAND_EXITED,
}
```
What do you think of that as a halfway house?

cartmouse: I've pushed an update to this which _kind of_ implements what I've said above. Would love to get your thoughts!

TF_Tooling_Web/src/Connection/TouchFreeServiceTypes.ts
cartmouse: Changed to an interface :+1:

TF_Service_dotNet/TouchFree/InteractionZoneEvent.cs
cartmouse: Updated to a record

CHANGELOG-windows.md
cartmouse: Not really sure. I don't mind removing it if it's not helpful, but I don't think there's any harm in it being there either 🤷

TF_Tooling_Web/src/Connection/TouchFreeServiceTypes.ts
Craig-J: Good compromise - personally I'd go for a mapped type with the string literal union to translate but this implementation addresses the main issue where state was being overloaded.

Craig-J (approved):

PR #433, opened 2023-02-02
[TF-1112]: Build web settings using Tauri on CI

JordanBerni-UL:
## Summary

Build the Web Settings using Tauri on CI. The output should be a signed .exe which runs the web settings at http://localhost:9739/settings (hosted by the service) and the build files (in dist) to use on BrightSign
https://ultrahaptics.atlassian.net/browse/TF-1112

With the addition of Tauri we can now access the filesystem so the hidden "Open Tracking Logs" and "Open TouchFree Logs" buttons have been re-added (but only when ran in the Tauri app)

### Tests Added

Updated Windows Specific test plan: 
https://ultrahaptics.atlassian.net/projects/TF?selectedItem=com.atlassian.plugins.atlassian-connect-plugin:com.kanoah.test-manager__main-project-page#!/testCase/TF-T439
https://ultrahaptics.atlassian.net/projects/TF?selectedItem=com.atlassian.plugins.atlassian-connect-plugin:com.kanoah.test-manager__main-project-page#!/testCase/TF-T440

## Contributor Tasks

- [x] PO review (optional depending on work type)

## Reviewer Tasks

- [ ] Developer testing
- [ ] Code reviewed


.gitlab-ci.yml
```diff
@@ -348,22 +348,33 @@ TF_Settings::Windows::build:
...
+  image: registry.ultrahaptics.com/ooh/infrastructure/windows-touchfree-toolchain:latest
   script:
     - npm -v
     - npm install -w TF_Settings_Web
     - npm ls
     - cd ./TF_Settings_Web
-    - npm run build
-    - npx generate-license-file --input ./package.json --output ./dist/ThirdPartyLicenses.txt --eol posix
+    - npm run tauri build
+    # npm workspaces cause the Web Settings node modules to be installed in the root folder so need to copy the Web Settings' package.json up a level
```
Craig-J, line 358: Why do we need to do this? Won't it overwrite the workspaces package.json? 🤔

TF_Settings_Web/src/TauriUtils.ts
```diff
@@ -0,0 +1,17 @@
+import { open } from '@tauri-apps/api/shell';
+import { appWindow } from '@tauri-apps/api/window';
+
+export const isDesktop = () => '__TAURI_METADATA__' in window;
```
Craig-J, line 6: Is this documented in Tauri docs? Could do with a comment explaining why/how this works.
👀

JordanBerni-UL: It's documented on their changelog 
`Breaking change: Move __currentWindow and __windows values from window.__TAURI__ to window.__TAURI_METADATA__`
but I found that solution via the forums and then console logging window while in a Tauri environment.

Will add a comment

.gitlab-ci.yml
JordanBerni-UL: We've already built the app and everything so it's fine to overwrite the other package.json as it's being used again (and docker image so won't affect other runs). 

We need to do it as TF_Settings_Web has no node modules (they get installed at the root directory) so generating the license gives you an empty file (has been broken since we did the workspaces change). For the license to correctly populate we need the web setting's package.json and the node modules at the same level so need to copy the package.json up a level to the root directory.
I added a comment trying to explain this but if you don't think it's clear then let me know what I missed

Also added a check to fail the job fi the license file doesn't contain React (as it definitely should if it populated correctly)

Craig-J: It's possible for a node_modules to get generated in the TF_Settings_Web directory - it will miss licenses from those with this setup. It's ok for now if that's not happening (I believe it happens with some kind of situation with ambiguous dependencies) but maybe it's worth putting a check in place to alert us if that starts happening? Something really basic like the job checks if a `node_modules` exists in TF_Settings_Web and warns us it's time to think about this.

Craig-J (approved):

PR #437, opened 2023-02-17
Bugfix/tf 1101 validate primitive types

Craig-J:
## Summary

Fix validation issue where service would crash when receiving non-integer value for screen height/width when setting config.

Closes [TF-1101](https://ultrahaptics.atlassian.net/browse/TF-1101)

PR includes some refactoring:

- `MessageQueueHandler`s now have an explicit validation step. Validation uses `Result<T>` to propagate errors and facilitate better error reporting. Default validation checks that `requestID` exists and further validation is optional for message handlers.
- Added `ValidatedIncomingRequest` which is produced from `IncomingRequest` by validation step. Removed `requestId` from `IncomingRequest` as it requires validation to produce.
- Added methods for sending basic success and error responses to `MessageQueueHandler`.
- Moved validation functions to new `MessageValidation` static class to facilitate testing. The intention here is to share common validation operations between handlers. A better solution may be adding validation to `IMessageQueueHandler` and combining message handlers that use the same message format but this is a much bigger refactoring than I wanted to handle here.
- Improvements to `Result<T>`
  - Added `Empty` type to be used Results that have no contained value but are not an error. Just for `Success` currently.
  - `Result<T>` is now a struct rather than abstract class with derived `Some<T>` and `Error<T>`.
  - Added `Error` type for creating `Result<T>` instances representing an error. Any other `T` is considered a successful result.
  - Added `Map` function which transforms a `Result<T>` into another `Result<TMapped>` given a mapping function, or propagates any errors.

### Tests Added

_Summary of the tests added as part of this change, or reason why no tests are required._

Added validation tests for `ValidateConfigJson` and moved tests for `ValidateRequestId` (previously `RequestIdExists`) to new `MessageValidationTests` file.

## Contributor Tasks

- [x] QA review (or another developer if no QA is available)
    - [x] [TouchFree BrightSign](/ultraleap/touchfree/blob/-/CHANGELOG-brightsign.md)

## Reviewer Tasks

- [x] Developer testing
- [x] Code reviewed

[TF-1101]: https://ultrahaptics.atlassian.net/browse/TF-1101?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


TF_Service_dotNet/TouchFree/Connections/MessageValidation.cs
```diff
@@ -0,0 +1,63 @@
...
+
+        /// <summary>
+        /// Checks for invalid states of the config request
+        /// </summary>
+        /// <param name="contentObj">JObject of content</param>
+        /// <returns>Returns a response as to the validity of the _content</returns>
+        public static Result<Empty> ValidateConfigJson(JObject contentObj)
+        {
+            bool anyValidContentFound = false;
+            foreach (var contentElement in contentObj)
```
GabrielHomewoodUL, line 36: 👀 This doesn't currently handle any unexpected elements in the config json. At the moment we would return an error response if there is additional content at any level that we are not expecting.

e.g. "touch" in the base level

TF_Service_dotNet/TouchFree/Connections/MessageValidation.cs
```diff
@@ -0,0 +1,63 @@
...
+                {
+                    try
+                    {
+                        // Check that _contentObj matches ConfigState structure, including all types
+                        JsonConvert.DeserializeObject(contentElement.Value.ToString(), T);
+                        anyValidContentFound = true;
+                    }
+                    catch (JsonReaderException ex)
+                    {
+                        return new Error("Setting configuration failed with message:\n\"" + ex.Message + "\"");
```
GabrielHomewoodUL: 🙈 Would be nice to use string interpolation here

TF_Service_dotNet/TouchFree/Connections/MessageValidation.cs
Craig-J: This is deliberate. From a backwards compatibility standpoint being less strict is beneficial as we can remove parts of the message structure without causing it to fail validation. Clients can continue to send data that is removed in newer versions of the protocol to maintain backwards compatibility if required.

The validation scheme I imagine is most flexible and would like to standardize on amongst message queues is:
- Check always required fields are present
- Check for optional fields
- Check contextual constraints e.g. "at least one in set of n fields is required", "if x exists y is also required", "if k exists j must not exist"
- Ignore unexpected fields

Alternatively we could version message queue handlers and be strict about unexpected fields. This is more flexible long term but a lot more work that I wouldn't want to take on right now.

TF_Service_dotNet/TouchFree/Connections/MessageValidation.cs
Craig-J: Fixed this

TF_Service_dotNet/TouchFree/Connections/MessageValidation.cs
Craig-J: After discussion last week - I've added functionality to fail validation when there's excess content fields in a config message and added a test case.

CHANGELOG-windows.md
```diff
@@ -41,7 +41,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Unity Settings: Cursor config correctly reloads when modified on disk
 - Service: Fixed an issue where the TouchFree Service would frequently throw exceptions when a client disconnects
 - Overlay Application: Fixed an issue where modified cursor alpha values would not be updated correctly
-
+- Service: Fixed an issue where the TouchFree Service would crash when non-integer configuration value was sent for an integer configuration property e.g. screen width.
```
GabrielHomewoodUL, line 44: This should also be added to the BrightSign configuration

TF_Service_dotNet/TouchFree/Connections/MessageQueues/MessageQueueHandler.cs
```diff
@@ -41,55 +40,39 @@ protected virtual void OnUpdate()
...
+            if (!queue.TryDequeue(out var content)) return;
+            HandleRequest(content);
         }
 
-        protected virtual void CreateAndSendNoRequestIdError(IncomingRequest _request)
-        {
-            ResponseToClient failureResponse = new ResponseToClient(string.Empty, "Failure", noRequestIdFailureMessage, _request.content);
+        private void HandleRequest(IncomingRequest request) =>
+            request.DeserializeAndValidateRequest(jObject => ValidateContent(jObject, request))
+                .Match(Handle, error => HandleValidationError(request, error));
```
GabrielHomewoodUL: With this version of error handling we never send back error messages with requestIDs set and so it will not trigger callbacks in tooling for those messages

GabrielHomewoodUL (approved):

PR #440, opened 2023-02-22
Improvements and fixes based on new web config changes

MattFiler:
## Summary

Fixes various issues discovered while investigating [TF-1138](https://ultrahaptics.atlassian.net/browse/TF-1138):
- Web config now saves more reliably
- Unity settings now respect new web config values
- Unity settings correctly reload visuals config
- Overlay app respects alpha value changes correctly

The root issue of TF-1138 still remains unfixed (seems to be system-specific), but figured it was useful to merge these fixes in the meantime to prevent further merge conflicts.

## Contributor Tasks

- [ ] Ensure documentation requirements are met e.g., public API is commented
- [ ] Consider any licensing/other legal implications e.g., notices required

- [ ] Make sure the fix version on the issue is set correctly

## Reviewer Tasks

- [x] Developer testing
- [x] Code reviewed

[TF-1138]: https://ultrahaptics.atlassian.net/browse/TF-1138?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG-windows.md
+++ b/CHANGELOG-windows.md
@@ -28,14 +28,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Service: `HandEntered` & `HandExited` events, sent when the active hand enters and exits the interaction zone (if enabled) respectively
 - Service: Removed informational level logging to improve readability of log files
+- Unity Settings: Added preset options for dark/light outline cursor
 
 ### Fixed
 
 - Unity Settings/Service: Fixed an issue where using Touch Plane with Scroll and Drag caused cursor jumps
 - Unity Settings/Service: Fixed an issue where clicking was very difficult while Touch Plane and Scroll and Drag were both active
 - Unity Settings: Fixed an issue with moving the bottom masking slider
+- Unity Settings: Cursor config correctly reloads when modified on disk
 - Service: Fixed an issue where the TouchFree Service would frequently throw exceptions when a client disconnects
-- Overlay Application: Fixed an issue where modified cursor alpha values would not be updated correctly.
+- Overlay Application: Fixed an issue where modified cursor alpha values would not be updated correctly
 
 
 ## [2.4.0] - 2022-10-14
--- a/CHANGELOG-windows.md
+++ b/CHANGELOG-windows.md
@@ -34,7 +34,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Unity Settings/Service: Fixed an issue where clicking was very difficult while Touch Plane and Scroll and Drag were both active
 - Unity Settings: Fixed an issue with moving the bottom masking slider
 - Service: Fixed an issue where the TouchFree Service would frequently throw exceptions when a client disconnects
-- Overlay Application: Fixed an issue where alpha values would not be updated within the same instance of a cursor when modifying through web settings.
+- Overlay Application: Fixed an issue where modified cursor alpha values would not be updated correctly.
 
 ## [2.4.0] - 2022-10-14
 
--- a/CHANGELOG-windows.md
+++ b/CHANGELOG-windows.md
@@ -34,6 +34,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Unity Settings/Service: Fixed an issue where clicking was very difficult while Touch Plane and Scroll and Drag were both active
 - Unity Settings: Fixed an issue with moving the bottom masking slider
 - Service: Fixed an issue where the TouchFree Service would frequently throw exceptions when a client disconnects
+- Overlay Application: Fixed an issue where alpha values would not be updated within the same instance of a cursor when modifying through web settings.
 
 ## [2.4.0] - 2022-10-14
 
--- a/TF_Settings_Web/src/Pages/Visuals/ColorPicker.tsx
+++ b/TF_Settings_Web/src/Pages/Visuals/ColorPicker.tsx
@@ -38,6 +38,11 @@ const ColorPicker: React.FC<ColorPickerProps> = ({ cursorStyle, updateCursorStyl
                     color={cursorStyle[activeTabIndex]}
                     onChange={(color) => updateCursorStyle({ ...cursorStyle, [activeTabIndex]: color })}
                     onPointerUp={() => writeOutConfig()}
+                    onPointerLeave={(e) => {
+                        if (e.pressure != 0) {
+                            writeOutConfig();
+                        }
+                    }}
                 />
                 <input
                     type="text"
--- a/TF_Settings_Web/src/Pages/Visuals/ColorPicker.tsx
+++ b/TF_Settings_Web/src/Pages/Visuals/ColorPicker.tsx
@@ -10,11 +10,12 @@ import { CursorStyle } from './VisualsUtils';
 interface ColorPickerProps {
     cursorStyle: CursorStyle;
     updateCursorStyle: (style: CursorStyle) => void;
+    writeVisualsConfigIfNew: () => void;
 }
 
 const cursorSections = ['Center Fill', 'Outer Fill', 'Center Border'] as const;
 
-const ColorPicker: React.FC<ColorPickerProps> = ({ cursorStyle, updateCursorStyle }) => {
+const ColorPicker: React.FC<ColorPickerProps> = ({ cursorStyle, updateCursorStyle, writeVisualsConfigIfNew }) => {
     const [activeTabIndex, setActiveTabIndex] = useState<number>(0);
 
     return (
@@ -41,7 +42,10 @@ const ColorPicker: React.FC<ColorPickerProps> = ({ cursorStyle, updateCursorStyl
                     type="text"
                     className={'color-picker__body__text'}
                     value={cursorStyle[activeTabIndex].toUpperCase()}
-                    onChange={(e) => updateCursorStyle({ ...cursorStyle, [activeTabIndex]: e.target.value })}
+                    onChange={(e) => {
+                        updateCursorStyle({ ...cursorStyle, [activeTabIndex]: e.target.value });
+                        writeVisualsConfigIfNew();
+                    }}
                 />
             </div>
         </div>
--- a/TF_Settings_Web/src/Pages/Visuals/ColorPicker.tsx
+++ b/TF_Settings_Web/src/Pages/Visuals/ColorPicker.tsx
@@ -10,12 +10,11 @@ import { CursorStyle } from './VisualsUtils';
 interface ColorPickerProps {
     cursorStyle: CursorStyle;
     updateCursorStyle: (style: CursorStyle) => void;
-    writeOutConfig: () => void;
 }
 
 const cursorSections = ['Center Fill', 'Outer Fill', 'Center Border'] as const;
 
-const ColorPicker: React.FC<ColorPickerProps> = ({ cursorStyle, updateCursorStyle, writeOutConfig }) => {
+const ColorPicker: React.FC<ColorPickerProps> = ({ cursorStyle, updateCursorStyle }) => {
     const [activeTabIndex, setActiveTabIndex] = useState<number>(0);
 
     return (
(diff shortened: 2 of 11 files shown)


TF_Settings_Web/src/Pages/Visuals/ColorPicker.tsx
```diff
@@ -37,19 +36,12 @@ const ColorPicker: React.FC<ColorPickerProps> = ({ cursorStyle, updateCursorStyl
...
                     color={cursorStyle[activeTabIndex]}
                     onChange={(color) => updateCursorStyle({ ...cursorStyle, [activeTabIndex]: color })}
-                    onPointerUp={() => writeOutConfig()}
                 />
                 <input
                     type="text"
                     className={'color-picker__body__text'}
                     value={cursorStyle[activeTabIndex].toUpperCase()}
                     onChange={(e) => updateCursorStyle({ ...cursorStyle, [activeTabIndex]: e.target.value })}
-                    onBlur={() => writeOutConfig()}
```
JordanBerni-UL, line 55: 👀 It's now possible to edit the custom color hex string then choose a different tab to leave the page without writing that update. 

Solution could be to always write out the config on navigating away from the page (DM if unsure about this)

TF_Settings_Web/src/Pages/Visuals/VisualsScreen.tsx
```diff
@@ -60,7 +60,8 @@ const VisualsScreen: React.FC = () => {
     const cursorSection = useRef<HTMLDivElement>(null);
 
     const [state, dispatch] = useReducer(reducer, defaultVisualsConfig);
-    const writeOutConfig = () => dispatch({ content: {}, writeOutConfig: true });
+    const updateConfig = (c = {}) => dispatch({ content: c, writeOutConfig: false });
```
JordanBerni-UL: 👍 I like this simplification

TF_Settings_Web/src/Pages/Visuals/VisualsScreen.tsx
```diff
@@ -92,7 +93,7 @@ const VisualsScreen: React.FC = () => {
     if (!isDesktop() || !hasReadConfig) return <></>;
 
     return (
-        <div className={classes('scroll-div')}>
+        <div className={classes('scroll-div')} onPointerUp={writeConfig}>
```
JordanBerni-UL: 💬 This is going to result in a lot more writes to the config (i.e. spam clicking the background will end up writing)
I don't think this is too big an issue

JordanBerni-UL: 👀 If we are only writing out in this place we could just call `writeVisualsConfig` here (instead of in the reducer) meaning we can change the reducer's signature to just `const reducer = (state: VisualsConfig, content: Partial<VisualsConfig>)` and then simplify the calls to it further (could just call `dispatch({key: value})` and remove updateConfig method)

TF_Settings_Web/src/Pages/Visuals/VisualsScreen.tsx
JordanBerni-UL: (was nice while it lasted - see below comment)

TF_Settings_Web/src/Pages/Visuals/VisualsScreen.tsx
MattFiler: Nice, added this

TF_Settings_Web/src/Pages/Visuals/ColorPicker.tsx
MattFiler: Changed to listen to `onpointerup` for the whole window now - should resolve this?

TF_Settings_Web/src/Pages/Visuals/VisualsScreen.tsx
```diff
@@ -47,11 +47,8 @@ const bgImages = [GradientBg, WhiteTextBg, BlackTextBg, MountainBg];
 const bgPreviewImages = [GradientBgPreview, WhiteTextBgPreview, BlackTextBgPreview, MountainBgPreview];
 const closeCtiOptions = ['Users Hand Present', 'User Performs Interaction'];
 
-const reducer = (state: VisualsConfig, payload: { content: Partial<VisualsConfig>; writeOutConfig: boolean }) => {
+const reducer = (state: VisualsConfig, payload: { content: Partial<VisualsConfig> }) => {
```
JordanBerni-UL: :eyes: You can take this out of payload now and just use `content: Partial<VisualsConfig>` - the payload wrapper was needed when there was multiple inputs (i.e. content and writeOutConfig) as the reducer only takes one (+ a parameter for the current state)

TF_Settings_Web/src/Pages/Visuals/VisualsScreen.tsx
```diff
@@ -47,11 +47,8 @@ const bgImages = [GradientBg, WhiteTextBg, BlackTextBg, MountainBg];
...
     const newState: VisualsConfig = { ...state, ...payload.content };
```
JordanBerni-UL: :eyes: can just return {...state, ...content} instead of assigning it first

TF_Settings_Web/src/Pages/Visuals/VisualsScreen.tsx
```diff
@@ -74,14 +70,21 @@ const VisualsScreen: React.FC = () => {
         section.setProperty('--center-border', cursorStyle[2]);
     };
 
+    let writtenConfig: VisualsConfig;
+    const writeVisualsConfigIfNew = () => {
+        if (writtenConfig == state) return;
```
JordanBerni-UL: :eyes: double equals is banned!!! Always triple

TF_Settings_Web/src/Pages/Visuals/VisualsScreen.tsx
```diff
@@ -74,14 +70,21 @@ const VisualsScreen: React.FC = () => {
         section.setProperty('--center-border', cursorStyle[2]);
     };
 
+    let writtenConfig: VisualsConfig;
```
JordanBerni-UL: :eyes: This would be nicer as a useRef rather than a floating variable

TF_Settings_Web/src/Pages/Visuals/VisualsScreen.tsx
```diff
@@ -91,6 +94,13 @@ const VisualsScreen: React.FC = () => {
 
     if (!isDesktop() || !hasReadConfig) return <></>;
 
+    window.onpointerup = () => {
```
JordanBerni-UL: :eyes: These should be set in a useEffect and we should register and unregister as window is global

CHANGELOG-windows.md
```diff
@@ -34,6 +34,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Unity Settings/Service: Fixed an issue where clicking was very difficult while Touch Plane and Scroll and Drag were both active
 - Unity Settings: Fixed an issue with moving the bottom masking slider
 - Service: Fixed an issue where the TouchFree Service would frequently throw exceptions when a client disconnects
+- Overlay Application: Fixed an issue where alpha values would not be updated within the same instance of a cursor when modifying through web settings.
```
JordanBerni-UL: :eyes: The Web Settings isn't release so wouldn't mention it (can mention that e.g. Overlay Application cursor now respects alpha values.... )

TF_Settings_Web/src/Pages/Visuals/VisualsScreen.tsx
JordanBerni-UL: It's still wrapped an extra layer (added seperate comment)

TF_Settings_Web/src/Pages/Visuals/VisualsScreen.tsx
```diff
@@ -47,23 +47,19 @@ const bgImages = [GradientBg, WhiteTextBg, BlackTextBg, MountainBg];
...
-};
-
 const VisualsScreen: React.FC = () => {
     const isLandscape = useIsLandscape();
     const cursorSection = useRef<HTMLDivElement>(null);
 
-    const [state, dispatch] = useReducer(reducer, defaultVisualsConfig);
-    const writeOutConfig = () => dispatch({ content: {}, writeOutConfig: true });
-    const [hasReadConfig, setHasReadConfig] = useState<boolean>(false);
+    const writtenConfig = useRef<VisualsConfig|null>(null);
```
JordanBerni-UL: 👀 If you just do `useRef<VisualsConfig>()` it implicitly is `useRef<VisualsConfig|undefined>()` when the initial value is empty so can do that instead of specifying null

TF_Settings_Web/src/Pages/Visuals/VisualsScreen.tsx
```diff
@@ -74,19 +70,30 @@ const VisualsScreen: React.FC = () => {
...
 
-    const currentPreset = useMemo((): CursorPreset => {
-        return presets[state.activeCursorPreset];
-    }, [state.activeCursorPreset]);
+    const updateConfig = (content: Partial<VisualsConfig>) => {
+        config.current = {...config.current, ...content};
+    };
+
+    const writeVisualsConfigIfNew = () => {
+        if (writtenConfig !== null && writtenConfig.current === config.current) return;
```
JordanBerni-UL: 👀 I don't think you need that first check as if writtenConfig.current is null then it will never equal config.current? (and if you do then can just do !writtenConfig which does a falsey check)

TF_Settings_Web/src/Pages/Visuals/VisualsScreen.tsx
```diff
@@ -160,17 +165,14 @@ const VisualsScreen: React.FC = () => {
...
-                                                tertiaryCustomColor: convertHexToRGBA(style[2]),
-                                            },
-                                            writeOutConfig: false,
+                                        updateConfig({
+                                            primaryCustomColor: convertHexToRGBA(style[0]),
+                                            secondaryCustomColor: convertHexToRGBA(style[1]),
+                                            tertiaryCustomColor: convertHexToRGBA(style[2]),
                                         });
                                         updateCursorPreview(style);
+                                        writeVisualsConfigIfNew();
```
JordanBerni-UL: 👀 This writes out the config every time the color picker is moved - we only want it to write out when it's finished (which I think the window.pointerup event listener should handle?)

TF_Settings_Web/src/Pages/Visuals/VisualsScreen.tsx
```diff
@@ -117,22 +124,20 @@ const VisualsScreen: React.FC = () => {
...
                                     selected={styleOptions.indexOf(currentPreset) ?? 0}
                                     options={styleOptions}
                                     onChange={(preset) => {
-                                        dispatch({
-                                            content: { activeCursorPreset: presets.indexOf(preset as CursorPreset) },
-                                            writeOutConfig: true,
-                                        });
+                                        updateConfig({ activeCursorPreset: presets.indexOf(preset as CursorPreset) });
                                         updateCursorPreview(cursorStyles[preset as CursorPreset]);
+                                        writeVisualsConfigIfNew();
```
JordanBerni-UL, line 149: 👀 I don't think this is needed

TF_Settings_Web/src/Pages/Visuals/VisualsScreen.tsx
JordanBerni-UL: To make it work on the hex input I think you would need to pass down the writeVisualsConfigIfNew to the ColorPicker and then onBlur / onKeyPress (enter key) call writeVisualsConfigIfNew. This should avoid writing it out loads

TF_Settings_Web/src/Pages/Visuals/VisualsScreen.tsx
MattFiler: I added this as otherwise the config is updated `onChange` which happens after the `pointerup` event fires, so you end up trying to write the non-updated config.

JordanBerni-UL: 👍 Just tested and see what you mean

TF_Settings_Web/src/Pages/Visuals/VisualsScreen.tsx
MattFiler: Oops, good catch! I did pass it through but then thought this solution was neater - was so focussed on the hex input that I forgot to check the sliders again 😅

TF_Settings_Web/src/Pages/Visuals/VisualsScreen.tsx
MattFiler: Nice - changed that

TF_Settings_Web/src/Pages/Visuals/VisualsScreen.tsx
MattFiler: Correct! I wasn't sure if `writtenConfig` was gonna be null on start but I should've checked 😅

GabrielHomewoodUL (approved):

PR #465, opened 2023-03-09
[TF-634]: InputActionManager Unit Test - plugin gets correct data

JordanBerni-UL:
## Summary

Add a unit test to check that InputActionManager calls the registered plugin with the correct data
https://ultrahaptics.atlassian.net/browse/TF-634

### Tests Added

This.

## Contributor Tasks

- [ ] Ensure documentation requirements are met e.g., public API is commented
- [ ] Consider any licensing/other legal implications e.g., notices required

## Reviewer Tasks

- [ ] Developer testing
- [ ] Code reviewed


TF_Tooling_Web/src/tests/testUtils.ts
```diff
@@ -15,3 +16,17 @@ export const mockTfInputAction = (input?: Partial<TouchFreeInputAction>) =>
             input?.ProgressToClick ?? 0
         )
     );
+
+export const mockTfPluginInputAction = (input?: Partial<TouchFreeInputAction>) =>
+    InputActionManager.HandleInputAction(
+        new TouchFreeInputAction(
```
cartmouse: :eyes: Looks like we might end up using this default structure frequently. Could it be functionalised to use across the two mocks?

TF_Tooling_Web/src/Plugins/tests/InputActionManager.test.ts
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    beforeAll(() => {
+        InputActionManager.SetPlugins([new MockPlugin()]);
+    });
+
+    test('Check plugin gets called with the correct data', () => {
+        expect(pluginCalled).toBeFalsy();
+        currentInputAction = new TouchFreeInputAction(
+            Date.now(),
```
cartmouse: :eyes: As per other comment, this could use that same function.

TF_Tooling_Web/src/Plugins/tests/InputActionManager.test.ts
```diff
@@ -0,0 +1,64 @@
...
+import { InputActionPlugin } from '../InputActionPlugin';
+
+describe('InputActionManager', () => {
+    let currentInputAction: TouchFreeInputAction;
+    let currentModifiedAction: TouchFreeInputAction;
+    let pluginCalled = false;
+
+    class MockPlugin extends InputActionPlugin {
+        RunPlugin(inputAction: TouchFreeInputAction): TouchFreeInputAction | null {
+            pluginCalled = true;
```
cartmouse: 💬 Would we want to do this after the following expect checks? So we know they've been run through?

JordanBerni-UL: Sure - will move it :)

TF_Tooling_Web/src/Plugins/tests/InputActionManager.test.ts
JordanBerni-UL: I've made a common function for the mock input action stuff but don't think this is useful here as I want to explicitly set the value for each parameter of the input action (to test they are all being set) and changing this to an object just for it to be unstructured and call the same constructor it does here seems overkill

cartmouse: I might be misunderstanding, but using your new `createInputAction` function here instead of the `new TouchFreeInputAction` only reduces code duplication and doesn't do any object assignment or destructuring?

JordanBerni-UL: Ah glossed over these were all the default parameters so can just call `createInputAction` without passing it anything

JordanBerni-UL: Updated

cartmouse (approved):

PR #474, opened 2023-03-17
[TF-1233]: Supress errors during snapshot test

JordanBerni-UL:
## Summary

When running TouchFreeToolingTypes.test.ts we get a lot of console errors.
This is expected as the snapshot test feeds every input combination to the flag functions and these functions log an error if the input doesn't contain the correct flag (which is a valid input).

We want to suppress these console errors during testing as it makes the test output a lot harder to read as there are a mass of console errors which don't actually affect the test. We store the errors in an array and print them out if any tests fail

https://ultrahaptics.atlassian.net/browse/TF-1233

## Contributor Tasks

## Reviewer Tasks

- [ ] Developer testing
- [ ] Code reviewed


TF_Tooling_Web/src/tests/TouchFreeToolingTypes.test.ts
```diff
@@ -65,6 +65,10 @@ const bitmaskFlagParams: BitmaskFlags[] = [
 ];
 
 describe('BitmaskFlag', () => {
+    // Suppress errors from console and store them in an array which we print only if a test fails
+    const errors: string[] = [];
+    console.error = jest.fn((msg: string) => errors.push(msg));
```
GabrielHomewoodUL: 🤷  Do we need to revert this so that we get other errors from other tests or is the environment reset on exiting the `describe`?

GabrielHomewoodUL (approved):

TF_Tooling_Web/src/tests/TouchFreeToolingTypes.test.ts
JordanBerni-UL: Fixed - the environment is set per test file it seems

PR #487, opened 2023-04-12
Add `SetCurrentCursor` to base TouchFree object

cartmouse:
## Summary

Adds the `SetCurrentCursor` function to the base of Web Tooling to allow users to add custom cursors to the TouchFree object.

## Contributor Tasks

- [ ] Relevant changelogs have been updated with user-visible changes
    - [x] [TouchFree Web Tooling](/ultraleap/touchfree/blob/-/TF_Tooling_Web/CHANGELOG.md)

## Reviewer Tasks

- [ ] Developer testing
- [x] Code reviewed

### DIFF
--- a/TF_Tooling_Web/src/tests/TouchFree.test.ts
+++ b/TF_Tooling_Web/src/tests/TouchFree.test.ts
@@ -1,4 +1,5 @@
 import { ConnectionManager } from '../Connection/ConnectionManager';
+import { SVGCursor } from '../Cursors/SvgCursor';
 import TouchFree from '../TouchFree';
 import { TouchFreeEventSignatures, TouchFreeEvent } from '../TouchFreeToolingTypes';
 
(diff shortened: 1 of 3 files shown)


TF_Tooling_Web/src/tests/TouchFree.test.ts
```diff
@@ -43,4 +44,11 @@ describe('TouchFree', () => {
         expect(ConnectionManager.iPAddress).toBe(newAddress.ip);
         expect(ConnectionManager.port).toBe(newAddress.port);
     });
+
+    it('Should set the cursor correctly', () => {
+        const cursor = new SVGCursor();
+        cursor.DisableCursor();
+        TouchFree.SetCurrentCursor(cursor);
+        expect(TouchFree.GetCurrentCursor()).toBe(cursor);
```
JordanBerni-UL, line 53: For completion sake could we test with DotCursor after this?

Can create one with empty images to make it easier:
`const dotCursor = new DotCursor(new Image(), new Image());`

cartmouse: Does this give us any extra coverage? I'm not sure this adds value to the test of this function

JordanBerni-UL: 🤷 My thinking was that SVGCursor is the default cursor type so testing with a different class might add robustness (don't know when this would ever hit that issue but don't see any harm in it). Also testing that subsequent calls sets the CurrentCursor correctly could be of use.

If you don't see any value then no need to add 🤷

cartmouse: The former point is why I added the disable line so they're definitely different. But I think you're right about the subsequent calls being important to capture. I'll add it in :+1:

JordanBerni-UL (approved):

